PR #53, opened 2025-02-11
Add support for struct type & corresponding operations

asl:
We are likely missing something like `struct_inject` for assigning a field for struct values, but this could be done later.

Lots of small fixes and refactorings while there.

### DIFF
--- a/test/Translate/Ops/struct.p4
+++ b/test/Translate/Ops/struct.p4
@@ -1,0 +1,111 @@
+// RUN: p4mlir-translate --typeinference-only %s | FileCheck %s
+
+struct P {
+    bit<32> f1;
+    bit<32> f2;
+}
+
+struct T {
+    int<32> t1;
+    int<32> t2;
+}
+
+struct S {
+    T s1;
+    T s2;
+}
+
+struct Empty {};
+
+// CHECK: !Empty = !p4hir.struct<"Empty">
+// CHECK: !PortId_t = !p4hir.struct<"PortId_t", _v: !b9i>
+// CHECK: !T = !p4hir.struct<"T", t1: !i32i, t2: !i32i>
+// CHECK: !S = !p4hir.struct<"S", s1: !T, s2: !T>
+// CHECK: !metadata_t = !p4hir.struct<"metadata_t", foo: !PortId_t>
+
+// CHECK-LABEL: module
+
+const T t = { 32s10, 32s20 };
+const S s = { { 32s15, 32s25}, t };
+
+// CHECK: %t = p4hir.const ["t"] #p4hir.aggregate<[#int10_i32i, #int20_i32i]> : !T
+// CHECK: %s = p4hir.const ["s"] #p4hir.aggregate<[#p4hir.aggregate<[#int15_i32i, #int25_i32i]> : !T, #p4hir.aggregate<[#int10_i32i, #int20_i32i]> : !T]> : !S
+
+const int<32> x = t.t1;
+const int<32> y = s.s1.t2;
+
+const int<32> w = .t.t1;
+
+// CHECK: %x = p4hir.const ["x"] #int10_i32i
+// CHECK: %y = p4hir.const ["y"] #int25_i32i
+// CHECK: %w = p4hir.const ["w"] #int10_i32i
+
+const T tt1 = s.s1;
+const Empty e = {};
+
+// CHECK: %tt1 = p4hir.const ["tt1"] #p4hir.aggregate<[#int15_i32i, #int25_i32i]> : !T
+// CHECK: %e = p4hir.const ["e"] #p4hir.aggregate<[]> : !Empty
+
+const T t1 = { 10, 20 };
+const S s1 = { { 15, 25 }, t1 };
+
+const int<32> x1 = t1.t1;
+const int<32> y1 = s1.s1.t2;
+
+const int<32> w1 = .t1.t1;
+
+const T t2 = s1.s1;
+
+struct PortId_t { bit<9> _v; }
+
+const PortId_t PSA_CPU_PORT = { _v = 9w192 };
+
+struct metadata_t {
+    PortId_t foo;
+}
+
+action test2(inout PortId_t port) {
+  port._v = port._v + 1; 
+}
+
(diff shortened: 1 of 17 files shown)


test/Translate/Ops/struct.p4
```diff
@@ -0,0 +1,111 @@
...
+
+struct S {
+    T s1;
+    T s2;
+}
+
+struct Empty {};
+
+// CHECK: !Empty = !p4hir.struct<"Empty">
+// CHECK: !PortId_t = !p4hir.struct<"PortId_t", _v: !b9i>
```
qobilidop, line 21: Maybe it's more readable to move `struct PortId_t { bit<9> _v; }` to before these checks?

test/Translate/Ops/struct.p4
```diff
@@ -0,0 +1,111 @@
...
+// CHECK: !T = !p4hir.struct<"T", t1: !i32i, t2: !i32i>
+// CHECK: !S = !p4hir.struct<"S", s1: !T, s2: !T>
+// CHECK: !metadata_t = !p4hir.struct<"metadata_t", foo: !PortId_t>
```
qobilidop, line 24: Same for `metadata_t`.

test/Translate/Ops/struct.p4
```diff
@@ -0,0 +1,111 @@
...
+
+// CHECK-LABEL: module
+
+const T t = { 32s10, 32s20 };
+const S s = { { 32s15, 32s25}, t };
+
+// CHECK: %t = p4hir.const ["t"] #p4hir.aggregate<[#int10_i32i, #int20_i32i]> : !T
```
qobilidop, line 31: I'm curious how this check works. In `struct.mlir` I see that `#int10_i32i = #p4hir.int<10> : !i32i` is manually defined. Is this automated in some way?

test/Translate/Ops/struct.p4
```diff
@@ -0,0 +1,111 @@
...
+
+struct metadata_t {
+    PortId_t foo;
+}
+
+action test2(inout PortId_t port) {
+  port._v = port._v + 1; 
+}
+
+// CHECK-LABEL: p4hir.func action @test2(%arg0: !p4hir.ref<!PortId_t> {p4hir.dir = #p4hir<dir inout>}) {
```
qobilidop, line 71: Not a request for this PR. Just wondering in general:

1. Is it possible to keep argument names like `port`? Maybe as a string attribute?
2. Is doing that useful? I was thinking maybe for debugging this could make the MLIR code more readable. But I'm not sure if there are any drawbacks.

test/Translate/Ops/struct.p4
```diff
@@ -0,0 +1,111 @@
...
+// Just few important bits here        
+action test(inout metadata_t meta) {
+   bit<9> vv;
+
+   PortId_t p1 = { _v = vv };
+
+   // CHECK: %[[VV_VAR:.*]] = p4hir.variable ["vv"] : <!b9i>
+   // CHECK: %[[VV_VAL:.*]] = p4hir.read %[[VV_VAR]] : <!b9i>
+   // CHECK: %[[STRUCT:.*]] = p4hir.struct (%[[VV_VAL]]) : !PortId_t
+   // CHECK: %[[P_VAR:.*]] = p4hir.variable ["p1", init] : <!PortId_t>
```
qobilidop, line 88: Just to clarify my understanding. Does `init` here mean this is a variable declaration with init value? But since the init value is assigned to the variable in a seperate `p4hir.assign` op below, I don't fully understand what difference there is w/ or w/o `init` here?

test/Translate/Ops/struct.p4
```diff
@@ -0,0 +1,111 @@
...
+            
+   PortId_t p;
+   bit<9> v;
+   v = p._v;
+
+   v = meta.foo._v;
+
+   meta.foo._v = 1;
+
+   // CHECK: p4hir.scope {
```
qobilidop, line 99: Just to clarify my understanding. Why does a function call require a new scope?

qobilidop (approved):
LGTM overall. Just have some nits and questions.

test/Translate/Ops/struct.p4
asl: MLIR allows one to define various aliases here and there. This includes:
 - Hints for SSA names
 - Aliases for attributes
 - Aliases for types

We do this to simplify the reading and to reduce the overall verboseness :)

test/Translate/Ops/struct.p4
asl: > * Is it possible to keep argument names like `port`? Maybe as a string attribute?

Note that SSA value name has no particular meaning, it could be removed, stripped out, etc. The argument names would need to be somehow preserved, yes. I decided to postpone this until we will have annotations in place as we'd need to support `@name` anyway :)

test/Translate/Ops/struct.p4
asl: Yes, as we're having in the description :) "init` means that there will be first use and that first use would be an assignment – essentially an initialization. Essentially this allows us to distinguish (should we need this):
```p4
bit<10> foo = 42;
```
from
```p4
bit<10> foo;
foo = 42;
```

test/Translate/Ops/struct.p4
asl: This is to correctly model copy-in / copy-out semantics for arguments. If we're having `out` / `inout` argument, then we'd need to:
 - Create a temporary
 - Copy the value into it (for `inout`)
 - Pass the reference to the temporary
 - Copy the value out
The scope is create to limit the lifetime of the call-related temporaries

test/Translate/Ops/struct.p4
qobilidop: Got. That makes sense to me now.

test/Translate/Ops/struct.p4
qobilidop: Interesting! Thanks for the explanation :)

asl: So, in the future we can certainly flatten the scope. But this would be a separate lowering step as otherwise we'd need to track the lifetime of temporaries until the very end. Also, if the function would be inlined, we'd just inline it into this scope and all the locals, etc. will be properly nested.

PR #66, opened 2025-02-13
Add support for headers

asl:


asl:
To give more context: the overall intention is to make implicit things explicit in the P4HIR. So we're having:
 - Explicit representation of validity bit
 - Header operations that affects validity are explicitly lowered to validity bit manipulation

Everything else is unified with structs

qobilidop (approved):
I think explicit representation of validity bit is a good idea!

PR #68, opened 2025-02-14
Add arithmetic shift operation

tagolog:


lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -136,6 +136,63 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+                                        "the signedness of the left-hand side operand";
+
+        auto actualResultWidth = resultBitsType.getWidth();
+        auto expectedResultWidth = lhsBitsType.getWidth();
+        if (actualResultWidth != expectedResultWidth)
+            return op->emitOpError() << "the resulting width of an arithmetic shift operation must "
+                                        "equal the width of the left-hand side operand";
+    }
+
+    if (::mlir::isa<P4HIR::BitsType>(rhsType)) {
```
asl, line 177: This could be simplified to:
```c++
if (auto rhsBits = mlir::dyn_cast<P4HIR::BitsType>(rhsType)) {
  if (rhsBits.isSigned())
           return op->emitOpError()
                   << "the right-hand side operand of an arithmetic shift must be unsigned";
}
```

Same in other places

tools/p4mlir-translate/translate.cpp
```diff
@@ -598,6 +604,16 @@ mlir::Value P4HIRConverter::emitConcatOp(const P4::IR::Concat *concatop) {
                                            getValue(concatop->right));
 }
 
+mlir::Value P4HIRConverter::emitShlOp(const P4::IR::Shl *op) {
```
asl, line 607: I do not think we need separate methods for this (at least here, in translator), just emit shift ops in `postorder` directly.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -136,6 +136,63 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+                                            ::mlir::Type rhsType, ::mlir::Type resultType) {
+    if (::mlir::isa<P4HIR::InfIntType>(lhsType)) {
+        if (!::mlir::isa<P4HIR::InfIntType>(rhsType))
+            // This is the error message from P4 frontend
+            return op->emitOpError()
+                   << "shift result type is arbitrary-precision int, but right operand is not "
+                      "constant; width of left operand of shift needs to be specified or both "
+                      "operands need to be constant";
+
+        if (!::mlir::isa<P4HIR::InfIntType>(resultType))
```
asl, line 153: I think we can start with the check that `resultType` equals to `lhsType`. This is a common invariant, so it could be hoisted out of conditions.

asl (changes requested):
See comments

include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
```diff
@@ -331,6 +331,72 @@ def ConcatOp : P4HIR_Op<"concat", [Pure]> {
...
+    %2 = p4hir.const #p4hir.int<3> : !p4hir.infint
+    %3 = p4hir.shl(%0 : !p4hir.bit<32>, %1 : !p4hir.bit<8>) : !p4hir.bit<32>
+    %4 = p4hir.shl(%0 : !p4hir.bit<32>, %2 : !p4hir.infint) : !p4hir.bit<32>
+    ```
+  }];
+
+  let results = (outs AnyIntP4Type:$result);
+  let arguments = (ins AnyIntP4Type:$lhs, AnyIntP4Type:$rhs);
+
+  // Custom builder to infer the result type with the proper width
```
asl, line 353: Looks like the comment was just copied from `concat` and does not make sense here. Also, I think you can use `TypesMatchWith` to match result type with lhs type.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
tagolog: Fixed

lib/Dialect/P4HIR/P4HIR_Ops.cpp
tagolog: Fixed

include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
tagolog: Fixed

tools/p4mlir-translate/translate.cpp
tagolog: Fixed

asl (approved):

PR #72, opened 2025-02-19
Improve binop operation.

tagolog:
Operation operands and result's type was changed to BitsType, MLIR test was changed to use operations on signed int operands.

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
@@ -276,14 +276,14 @@ def BinOp : P4HIR_Op<"binop", [Pure,
     should be the same.
 
     ```mlir
-    %7 = p4hir.binop(add, %1, %2) : !p4hir.bit<32>
-    %7 = p4hir.binop(mul, %1, %2) : !p4hir.bit<32>
+    %3 = p4hir.binop(add, %1, %2) : !p4hir.bit<32>
+    %4 = p4hir.binop(mul, %1, %2) : !p4hir.bit<32>
     ```
   }];
 
-  let results = (outs BitsType:$result);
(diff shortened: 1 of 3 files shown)


include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
```diff
@@ -281,17 +281,13 @@ def BinOp : P4HIR_Op<"binop", [Pure,
     ```
   }];
 
-  // TODO: get more accurate than AnyP4Type
-  let results = (outs AnyP4Type:$result);
+  let results = (outs BitsType:$result);
   let arguments = (ins Arg<BinOpKind, "binop kind">:$kind,
-                       AnyP4Type:$lhs, AnyP4Type:$rhs);
+                       BitsType:$lhs, BitsType:$rhs);
```
asl, line 286: I do not think this is a correct constraint. https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-varint-ops allows lots of binary operations on `Type_InfInt` (aka arbitrary precision integer) values. Indeed, all of them are required to be compile-time constants and therefore in many cases are just const folded before we are seeing them.

However, as we are modeling spec in the dialect we should not rely on this. It should be possible to write, say, `p4hir.binop(add, %a, %b)` where `a` and `b` will be of `p4hir.infint` type. Folding / canonicalization of them is a separate task.

So, long story short: here we need to allow both `BitsType` and `InfInt` types.

asl (changes requested):

include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
tagolog: Fixed

asl (approved):
Great, thanks!

PR #82, opened 2025-02-22
Implement varbit data type (without operations)

tagolog:
Signed-off-by: Vladimir Shiryaev <w@tagolog.com>

PR #83, opened 2025-02-24
Implement varbit data type (without operations)

tagolog:


qobilidop (approved):
Does this supercede #82? In that case could you close #82?

include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
```diff
@@ -88,6 +88,22 @@ def InfIntType : P4HIR_Type<"InfInt", "infint"> {
...
+//===----------------------------------------------------------------------===//
+
+def VarBitsType : P4HIR_Type<"VarBits", "varbit"> {
+  let summary = "bit-strings of dynamically-computed width with a fixed maximum width";
+  let description = [{
+    A variable-width bit-string type with a fixed maximum width.
+    The actual width is determined at runtime but cannot exceed the maximum.
+  }];
+
+  let parameters = (ins "unsigned":$maxWidth);
```
asl, line 103: Are we missing checks here?
 - varbit could be only in headers
 - it should be the last field (before validity)

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
 }
 
-LogicalResult HeaderType::verify(function_ref<InFlightDiagnostic()> emitError, StringRef,
+LogicalResult HeaderType::verify(function_ref<InFlightDiagnostic()> emitError, StringRef /*unused*/,
                                  ArrayRef<FieldInfo> elements) {
-    llvm::SmallDenseSet<StringAttr> fieldNameSet;
+
     LogicalResult result = success();
+    if (elements.empty()) {
+        result = failure();
```
asl, line 292: The only reason why this convoluted logic with `result` was used is to report all duplicate fields (not just exit on first error). Here everything is empty, so just do early return.

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
                         << "' in p4hir.header type";
         }
+    }
 
-    if (elements.empty()) emitError() << "empty p4hir.header type";
+    auto lastField = elements.back();
+    if (lastField.name != validityBit ||
+        !mlir::isa<P4HIR::ValidBitType>(lastField.type)) {
+            result = failure();
+            emitError() << "the last field of p4hir.header type should be validity bit " << lastField.name.getValue() << " " << lastField.type.getImpl()->getAbstractType().getName();
```
asl, line 310: 1. Just print type, e.g. `<< lastField.type`
2. The error message is misleading, could probably be
```c++
emitError() << "the last field of p4hir.header type should be validity bit, but got" << lastField.name.getValue() << " of type " << lastField.type;
```

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+    }
 
-    if (elements.back().name != validityBit ||
-        !mlir::isa<P4HIR::ValidBitType>(elements.back().type))
-        emitError() << "last field of p4hir.header type should be validity bit";
+    // The only varbit field can present in the header
```
asl, line 313:
```suggestion
    // Only varbit field can present in the header
```

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+    auto i = 0;
+    size_t varbitFieldIndex = std::numeric_limits<size_t>::max();
```
asl, line 315: Maybe you can just use `llvm::count_if` (counting # of fields of varbit type) and error out if result it more than `1`?

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+                    << "' in p4hir.header type";
+                break;
+            }
+            varbitFieldIndex = i;
+        }
+        i++;
+    }
+
+    // If a varbit field is found, ensure it is the last “data” field:
+    if (varbitFieldIndex != std::numeric_limits<size_t>::max()) {
```
asl, line 330: Replace with count check.

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+        size_t expectedIndex = elements.size() - 2;
+        if (varbitFieldIndex != expectedIndex) {
+            result = failure();
+            emitError() << "varbit field name '" << elements[varbitFieldIndex].name.getValue() << "' must be be immediately before the validity field in p4hir.header type";
+        }
+    }
+
+    // Helper function to build the nested ternary operations recursively
```
asl, line 338: Bogus comment?

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+            for (const auto &elt : elements) {
+                // varbit and validity bit are allowed at the top (header) level only
+                if (nestingLevel == 0 && mlir::isa<P4HIR::VarBitsType, P4HIR::ValidBitType>(elt.type)) {
+                    continue;
+                }
+                if (mlir::isa<P4HIR::BitsType, P4HIR::BoolType, P4HIR::SerEnumType>(elt.type)) {
+                    continue;
+                }
+                if (const auto structType = mlir::dyn_cast<P4HIR::StructType>(elt.type)) {
+                    checkFieldTypes(structType.getElements(), nestingLevel + 1);
```
asl, line 350: Use `getInnerTypes()`

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -284,25 +284,79 @@ LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+        size_t expectedIndex = elements.size() - 2;
+        if (varbitFieldIndex != expectedIndex) {
+            result = failure();
+            emitError() << "varbit field name '" << elements[varbitFieldIndex].name.getValue() << "' must be be immediately before the validity field in p4hir.header type";
+        }
+    }
+
+    // Helper function to build the nested ternary operations recursively
+    std::function<void(ArrayRef<FieldInfo>, int)> checkFieldTypes =
+        [&](ArrayRef<FieldInfo> elements, int nestingLevel) -> void {
```
asl, line 340: why `int`?

asl (changes requested):
See comments

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: Fixed to return immediately.

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: Done.

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: Changed to "only one varbit field is allowed in p4hir.header type"

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: Done.

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: Done.

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: Yes. The comment was adjusted to "Helper function to verify inner types recursively, inspecting nested structs"

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: The function name was changed to "verifyInnerTypes". It seems to me that "verify" verb correlates better with the outer method
LogicalResult HeaderType::verify(...

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: The parameter was changed to "bool isHeader".
Thus the initial call fo header is "verifyInnerTypes(elements, true);"
and the interanal call for struct is "verifyInnerTypes(structType.getElements(), false);"

asl (approved):

PR #86, opened 2025-02-26
First cut of parser support

asl:
There is no support for constructor arguments and parser apply's, otherwise should be pretty complete (except value sets). Masks and ranges are supported.


asl:
Parser control flow is made explicit:
 - There is always transition to `start` case emitted, it is not special name anymore
 - Parser state names are fully qualified (so when e.g. inlining one should not bother about renaming)
 - Accept / reject states are not special anymore. The packet status is signaled via special `parser_accept` / `parser_reject` op. The latter could optionally take an error (used to support `verify`)
 - `verify` calls are lowered to normal control flow and `parser_reject` op with error argument
 - Default reject cases in selects are explicitly added
 - Set operations are explicit as well

asl:
After https://github.com/p4lang/p4mlir-incubator/pull/87 it seems we're only missing `values_set`'s

qobilidop (approved):

Issue #93, opened 2025-03-04
How to model packet_in/packet_out and other externs from core.p4 in P4HIR?


qobilidop:
@asl brought up this discussion on Zulip. There seem to be several options. I'd like to summarize these options here for further discussion:

- Option 1: Model `packet_in`/`packet_out` the same as other blackbox externs in P4HIR. Leave it up to backend (or backend dialect) to deal with their semantics.
- Option 2: Model `packet_in`/`packet_out` semantics explicitly as builtin types and ops in P4HIR.
- Option 3: We can probably introduce another P4MIR dialect in which `packet_in`/`packet_out` semantics are modeled explicitly as in option 2, but go with option 1 in P4HIR.

My personal preference is option 2 or 3. Because certain optimizations are possible only if `packet_in`/`packet_out`'s semantics are known. For example, we could have an optional optimization pass to eliminate unnecessary `packet_in.extract` if there is no use of the extracted header or field in the rest of the P4 program. For general externs, without knowing its semantics, we cannot perform such optimization.

Another argument is that since the semantics for externs from `core.p4` are specified as part of the P4 language, these semantics should also be modeled explicitly in the core IR.

For option 3, the main benefit is that a backend can choose to lower from P4HIR or P4MIR depending on how much optimization it can reuse. But I personally don't have any particular backend in mind that can justify the split of P4HIR and P4MIR.

asl:
I would say that we won't be able to model proper validity bit semantics w/o special handling of `packet_in` – `extract` makes header valid.

qobilidop:
> I would say that we won't be able to model proper validity bit semantics w/o special handling of `packet_in` – `extract` makes header valid.

Good point! Agree with that.

fruffy:
>For example, we could have an optional optimization pass to eliminate unnecessary packet_in.extract if there is no use of the extracted header or field in the rest of the P4 program

Note that an extract call also moves the parser pointer ahead. So removing it can lead to incorrect behavior. But we can substitute these calls with `advance`.

asl:
I'm suggesting to introduce new small dialect – p4corelib that would include operations from core.p4. In particular:

1. `packet_in` and `packet_out` types
2. `PacketExtract` (two variants), `PacketLookahead`, `PacketAdvance` and `PacketLength` operations that would essentially represent methods of present `packet_in` extern
3. `PacketEmit` operation
4. `verify` would be lowered to the explicit control flow like now
5. `static_assert` would be constfolded.

PR #94, opened 2025-03-05
Add support for externs. Ensure we can call specialized functions / methods

asl:
Some fixes while there.

### DIFF
--- a/test/Translate/Parser/packet_in.p4
+++ b/test/Translate/Parser/packet_in.p4
@@ -1,0 +1,72 @@
+// RUN: p4mlir-translate --typeinference-only %s | FileCheck %s
+
+extern packet_in {
+    void extract<T>(out T hdr);
+    void extract<T>(out T variableSizeHeader,
+                    in bit<32> variableFieldSizeInBits);
+    T lookahead<T>();
+    void advance(in bit<32> sizeInBits);
+    bit<32> length();
+}
+
+// CHECK:  p4hir.extern @packet_in {
+// CHECK:    p4hir.overload_set @extract {
+// CHECK:      p4hir.func @extract_0<!p4hir.type_var<"T">>(!p4hir.ref<!p4hir.type_var<"T">> {p4hir.dir = #out})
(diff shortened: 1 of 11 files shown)


test/Translate/Parser/packet_in.p4
```diff
@@ -0,0 +1,72 @@
+// RUN: p4mlir-translate --typeinference-only %s | FileCheck %s
+
+extern packet_in {
+    void extract<T>(out T hdr);
+    void extract<T>(out T variableSizeHeader,
+                    in bit<32> variableFieldSizeInBits);
+    T lookahead<T>();
+    void advance(in bit<32> sizeInBits);
+    bit<32> length();
+}
```
qobilidop, line 10: Not a request to change this PR.

Now I'm confused whether we can tell a user-defined `packet_in` extern like this, from a `packet_in` extern from `core.p4`? I guess we can't?

Then I wonder how could we treat `packet_in` from `core.p4` differently as discussed in #93?

asl: Unfortunately, we cannot. As `core.p4` is included via preprocessor. So we can try somehow to track things to `p4include` dir, but this is pretty error prone. I would probably suggest us to invent some kind of annotation (e.g. `@system` or `@core`) to mark this in the future.

qobilidop: Completely agree!

fruffy: The inability to distiniguish system entities from user-defined ones is a long-standing problem in the compiler. Currently, we have the hacky solution where we check the include path, but that can be ambiguous based on user-configuration. 

I am thinking an explicit annotation might be the only way out... 

Maybe a particular special construct defined in the spec can work, too.

asl: > The inability to distiniguish system entities from user-defined ones is a long-standing problem in the compiler. Currently, we have the hacky solution where we check the include path, but that can be ambiguous based on user-configuration.

Even more, the behavior might be different with preprocessed sources (!)

PR #96, opened 2025-03-05
Implement header union type and header union type isValid built-in method

tagolog:

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
@@ -321,13 +321,6 @@ class StructLikeType<string name, string typeMnemonic>
       return getElements();
     }
 
-    template <typename Func>
-    void forEachField(Func &&func) const {
-      for (const auto &field : getFields()) {
-        func(field);
-      }
-    }
-
     void getInnerTypes(mlir::SmallVectorImpl<mlir::Type> &types) {
       for (const auto &field : getElements()) types.push_back(field.type);
     }
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
@@ -317,8 +317,15 @@ class StructLikeType<string name, string typeMnemonic>
        return {};
     }
 
-    llvm::ArrayRef<P4HIR::FieldInfo> getFields() {
+    llvm::ArrayRef<P4HIR::FieldInfo> getFields() const {
       return getElements();
+    }
+
+    template <typename Func>
+    void forEachField(Func &&func) const {
+      for (const auto &field : getFields()) {
+        func(field);
+      }
     }
 
     void getInnerTypes(mlir::SmallVectorImpl<mlir::Type> &types) {
@@ -480,14 +487,6 @@ def HeaderUnionType : StructLikeType<"HeaderUnion", "header_union"> {
     A header union represents an alternative containing at most one of several different headers.
     !p4hir.header_union<"name", field1: HeaderType1, field2: HeaderType2>
   }];
-
-  // We skip default builders entirely to consistently add validity bit field on fly
-  let skipDefaultBuilders = 1;
-
-  let builders = [
-    TypeBuilder<(ins "llvm::StringRef":$name,
-                      "llvm::ArrayRef<P4HIR::FieldInfo>":$fields)>
-  ];
 }
 
 //===----------------------------------------------------------------------===//
@@ -649,7 +648,7 @@ def AnyP4Type : AnyTypeOf<[BitsType, BooleanType, InfIntType, StringType,
 def AnyIntP4Type : AnyTypeOf<[BitsType, InfIntType]> {}
 def CallResultP4Type : AnyTypeOf<[BitsType, BooleanType, InfIntType, VoidType]> {}
 def LoadableP4Type : AnyTypeOf<[BitsType, BooleanType, InfIntType,
-                                StructType, HeaderType, Builtin_Tuple,
+                                StructType, HeaderType, HeaderUnionType, Builtin_Tuple,
                                 EnumType, SerEnumType, ErrorType,
                                 ValidBitType]> {}
 def AnyEnumType : AnyTypeOf<[EnumType, SerEnumType]>;
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
@@ -470,6 +470,27 @@ def HeaderType : StructLikeType<"Header", "header"> {
 }
 
 //===----------------------------------------------------------------------===//
+// HeaderUnionType
+//===----------------------------------------------------------------------===//
+
+// A header union. Declares the P4HIR::HeaderUnionType in C++.
+def HeaderUnionType : StructLikeType<"HeaderUnion", "header_union"> {
+  let summary = "header union type";
+  let description = [{
+    A header union represents an alternative containing at most one of several different headers.
+    !p4hir.header_union<"name", field1: HeaderType1, field2: HeaderType2>
+  }];
+
+  // We skip default builders entirely to consistently add validity bit field on fly
+  let skipDefaultBuilders = 1;
+
+  let builders = [
+    TypeBuilder<(ins "llvm::StringRef":$name,
+                      "llvm::ArrayRef<P4HIR::FieldInfo>":$fields)>
+  ];
+}
+
+//===----------------------------------------------------------------------===//
 // EnumType, ErrorType  & SerEnumType
 //===----------------------------------------------------------------------===//
 def EnumType : P4HIR_Type<"Enum", "enum", []> {
@@ -620,7 +641,7 @@ def ExternType : P4HIR_Type<"Extern", "extern", []> {
 //===----------------------------------------------------------------------===//
 
 def AnyP4Type : AnyTypeOf<[BitsType, BooleanType, InfIntType, StringType,
-                           StructType, HeaderType, Builtin_Tuple,
+                           StructType, HeaderType, HeaderUnionType, Builtin_Tuple,
                            EnumType, SerEnumType,
                            ValidBitType,
                            DontcareType, ErrorType, UnknownType,
@@ -632,7 +653,7 @@ def LoadableP4Type : AnyTypeOf<[BitsType, BooleanType, InfIntType,
                                 EnumType, SerEnumType, ErrorType,
                                 ValidBitType]> {}
 def AnyEnumType : AnyTypeOf<[EnumType, SerEnumType]>;
-def StructLikeType : AnyTypeOf<[StructType, HeaderType]>;
+def StructLikeType : AnyTypeOf<[StructType, HeaderType, HeaderUnionType]>;
 def P4ObjectType : AnyTypeOf<[ParserType, ExternType]>;
 
 /// A ref type with the specified constraints on the nested type.
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
@@ -317,7 +317,7 @@ class StructLikeType<string name, string typeMnemonic>
        return {};
     }
 
-    llvm::ArrayRef<P4HIR::FieldInfo> getFields() const {
+    llvm::ArrayRef<P4HIR::FieldInfo> getFields() {
       return getElements();
     }
 
--- a/lib/Dialect/P4HIR/P4HIR_Types.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Types.cpp
@@ -348,12 +348,6 @@ HeaderType HeaderType::get(mlir::MLIRContext *context, llvm::StringRef name,
     return Base::get(context, name, realFields);
 }
 
-HeaderUnionType HeaderUnionType::get(mlir::MLIRContext *context, llvm::StringRef name,
-                                     llvm::ArrayRef<FieldInfo> fields) {
-    llvm::SmallVector<FieldInfo, 4> realFields(fields);
-    return Base::get(context, name, realFields);
-}
-
 Type EnumType::parse(AsmParser &p) {
     std::string name;
     llvm::SmallVector<Attribute> fields;
--- a/lib/Dialect/P4HIR/P4HIR_Types.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Types.cpp
@@ -263,6 +263,13 @@ Type HeaderType::parse(AsmParser &p) {
     return Base::get(p.getContext(), name, parameters);
 }
 
+Type HeaderUnionType::parse(AsmParser &p) {
+    llvm::SmallVector<FieldInfo, 4> parameters;
+    std::string name;
+    if (parseFields(p, name, parameters)) return {};
+    return get(p.getContext(), name, parameters);
+}
+
 LogicalResult StructType::verify(function_ref<InFlightDiagnostic()> emitError, StringRef,
                                  ArrayRef<FieldInfo> elements) {
     llvm::SmallDenseSet<StringAttr> fieldNameSet;
@@ -300,8 +307,37 @@ LogicalResult HeaderType::verify(function_ref<InFlightDiagnostic()> emitError, S
     return result;
 }
 
+LogicalResult HeaderUnionType::verify(function_ref<InFlightDiagnostic()> emitError, StringRef,
+                                      ArrayRef<FieldInfo> elements) {
+    llvm::SmallDenseSet<StringAttr> fieldNameSet;
+    LogicalResult result = success();
+    fieldNameSet.reserve(elements.size());
+    for (const auto &elt : elements) {
+        if (!fieldNameSet.insert(elt.name).second) {
+            result = failure();
+            emitError() << "duplicate field name '" << elt.name.getValue()
+                        << "' in p4hir.header_union type";
+        }
+
+        // Check that all elements are header types
+        if (!mlir::isa<HeaderType>(elt.type)) {
+            result = failure();
+            emitError() << "header_union field '" << elt.name.getValue()
+                        << "' must be a header type";
+        }
+    }
+
+    if (elements.empty()) {
+        emitError() << "empty p4hir.header_union type";
+        return failure();
+    }
+
+    return result;
+}
+
 void StructType::print(AsmPrinter &p) const { printFields(p, getName(), getElements()); }
 void HeaderType::print(AsmPrinter &p) const { printFields(p, getName(), getElements()); }
+void HeaderUnionType::print(AsmPrinter &p) const { printFields(p, getName(), getElements()); }
 
 HeaderType HeaderType::get(mlir::MLIRContext *context, llvm::StringRef name,
                            llvm::ArrayRef<FieldInfo> fields) {
@@ -309,6 +345,12 @@ HeaderType HeaderType::get(mlir::MLIRContext *context, llvm::StringRef name,
     realFields.push_back(
         {mlir::StringAttr::get(context, validityBit), P4HIR::ValidBitType::get(context)});
 
+    return Base::get(context, name, realFields);
+}
+
+HeaderUnionType HeaderUnionType::get(mlir::MLIRContext *context, llvm::StringRef name,
+                                     llvm::ArrayRef<FieldInfo> fields) {
+    llvm::SmallVector<FieldInfo, 4> realFields(fields);
     return Base::get(context, name, realFields);
 }
 
--- a/test/Translate/Ops/header_union.p4
+++ b/test/Translate/Ops/header_union.p4
@@ -1,0 +1,54 @@
+// RUN: p4mlir-translate --typeinference-only %s | FileCheck %s
+
+header H1 {
+  bit<8> f;
+}
+
+header H2 {
+  bit<16> g;
+}
+
+header_union U {
+  H1 h1;
+  H2 h2;
+}
+
+action header_union_isValid_test() {
+  U u;
+  if (u.isValid())
+      return;
+}
+
+// CHECK: #[[$ATTR_0:.+]] = #p4hir.bool<false> : !p4hir.bool
+// CHECK: #[[$ATTR_1:.+]] = #p4hir.bool<true> : !p4hir.bool
+// CHECK: #[[$ATTR_2:.+]] = #p4hir<validity.bit valid> : !validity_bit
+// CHECK-LABEL:   p4hir.func action @header_union_isValid_test() {
+// CHECK:           %[[VAL_0:.*]] = p4hir.variable ["u"] : <!p4hir.header_union<"U", h1: !H1_, h2: !H2_>>
+// CHECK:           %[[VAL_1:.*]] = p4hir.const #[[$ATTR_0]]
+// CHECK:           %[[VAL_2:.*]] = p4hir.struct_extract_ref %[[VAL_0]]["h1"] : <!p4hir.header_union<"U", h1: !H1_, h2: !H2_>>
+// CHECK:           %[[VAL_3:.*]] = p4hir.struct_extract_ref %[[VAL_2]]["__valid"] : <!H1_>
+// CHECK:           %[[VAL_4:.*]] = p4hir.read %[[VAL_3]] : <!validity_bit>
+// CHECK:           %[[VAL_5:.*]] = p4hir.const #[[$ATTR_2]]
+// CHECK:           %[[VAL_6:.*]] = p4hir.cmp(eq, %[[VAL_4]], %[[VAL_5]]) : !validity_bit, !p4hir.bool
+// CHECK:           %[[VAL_7:.*]] = p4hir.ternary(%[[VAL_1]], true {
+// CHECK:             %[[VAL_8:.*]] = p4hir.const #[[$ATTR_1]]
+// CHECK:             p4hir.yield %[[VAL_8]] : !p4hir.bool
+// CHECK:           }, false {
+// CHECK:             p4hir.yield %[[VAL_6]] : !p4hir.bool
+// CHECK:           }) : (!p4hir.bool) -> !p4hir.bool
+// CHECK:           %[[VAL_9:.*]] = p4hir.struct_extract_ref %[[VAL_0]]["h2"] : <!p4hir.header_union<"U", h1: !H1_, h2: !H2_>>
+// CHECK:           %[[VAL_10:.*]] = p4hir.struct_extract_ref %[[VAL_9]]["__valid"] : <!H2_>
+// CHECK:           %[[VAL_11:.*]] = p4hir.read %[[VAL_10]] : <!validity_bit>
+// CHECK:           %[[VAL_12:.*]] = p4hir.const #[[$ATTR_2]]
+// CHECK:           %[[VAL_13:.*]] = p4hir.cmp(eq, %[[VAL_11]], %[[VAL_12]]) : !validity_bit, !p4hir.bool
+// CHECK:           %[[VAL_14:.*]] = p4hir.ternary(%[[VAL_7]], true {
+// CHECK:             %[[VAL_15:.*]] = p4hir.const #[[$ATTR_1]]
+// CHECK:             p4hir.yield %[[VAL_15]] : !p4hir.bool
+// CHECK:           }, false {
+// CHECK:             p4hir.yield %[[VAL_13]] : !p4hir.bool
+// CHECK:           }) : (!p4hir.bool) -> !p4hir.bool
+// CHECK:           p4hir.if %[[VAL_14]] {
+// CHECK:             p4hir.return
+// CHECK:           }
+// CHECK:           p4hir.return
+// CHECK:         }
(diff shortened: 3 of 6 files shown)


include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
```diff
@@ -401,6 +401,27 @@ def HeaderType : StructLikeType<"Header", "header"> {
...
+
+// A header union. Declares the P4HIR::HeaderUnionType in C++.
+def HeaderUnionType : StructLikeType<"HeaderUnion", "header_union"> {
+  let summary = "header union type";
+  let description = [{
+    A header union represents an alternative containing at most one of several different headers.
+    !p4hir.header_union<"name", field1: HeaderType1, field2: HeaderType2>
+  }];
+
+  // We skip default builders entirely to consistently add validity bit field on fly
```
asl, line 416: Do we really need this? One can construct union from headers only that already contain validity bits.

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -264,6 +300,12 @@ HeaderType HeaderType::get(mlir::MLIRContext *context, llvm::StringRef name,
     return Base::get(context, name, realFields);
 }
 
+HeaderUnionType HeaderUnionType::get(mlir::MLIRContext *context, llvm::StringRef name,
+                                     llvm::ArrayRef<FieldInfo> fields) {
+    llvm::SmallVector<FieldInfo, 4> realFields(fields);
```
asl, line 305: This looks strange. Why copy fields?

test/Translate/Ops/header_union.p4
```diff
@@ -0,0 +1,54 @@
...
+}
+
+header_union U {
+  H1 h1;
+  H2 h2;
+}
+
+action header_union_isValid_test() {
+  U u;
+  if (u.isValid())
```
asl, line 18: You also need tests for `#` literal as well as `setValid` / `setInvalid`, etc. calls

test/Translate/Ops/header_union.p4
```diff
@@ -0,0 +1,54 @@
...
+// CHECK: #[[$ATTR_2:.+]] = #p4hir<validity.bit valid> : !validity_bit
+// CHECK-LABEL:   p4hir.func action @header_union_isValid_test() {
+// CHECK:           %[[VAL_0:.*]] = p4hir.variable ["u"] : <!p4hir.header_union<"U", h1: !H1_, h2: !H2_>>
+// CHECK:           %[[VAL_1:.*]] = p4hir.const #[[$ATTR_0]]
+// CHECK:           %[[VAL_2:.*]] = p4hir.struct_extract_ref %[[VAL_0]]["h1"] : <!p4hir.header_union<"U", h1: !H1_, h2: !H2_>>
+// CHECK:           %[[VAL_3:.*]] = p4hir.struct_extract_ref %[[VAL_2]]["__valid"] : <!H1_>
+// CHECK:           %[[VAL_4:.*]] = p4hir.read %[[VAL_3]] : <!validity_bit>
+// CHECK:           %[[VAL_5:.*]] = p4hir.const #[[$ATTR_2]]
+// CHECK:           %[[VAL_6:.*]] = p4hir.cmp(eq, %[[VAL_4]], %[[VAL_5]]) : !validity_bit, !p4hir.bool
+// CHECK:           %[[VAL_7:.*]] = p4hir.ternary(%[[VAL_1]], true {
```
asl, line 33: The generated code looks a bit weird, as we're evaluating all nested `isValid`'s even if we already deduced that the the result is true. I think we'd need to shortcut and emit something like this:
`(H.h1.isValid() ? true : (H.h2.isValid() ? true : (H.h3.isValid() ? : ...)))`. So, it will be a sequence of nested ternary operators where you're yielding `true` in `true` region and recurse in `false` region.

tools/p4mlir-translate/translate.cpp
```diff
@@ -329,6 +330,10 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
         // Should be resolved eslewhere
         return false;
     }
+    bool preorder(const P4::IR::InvalidHeaderUnion *hu) override {
```
asl, line 333: We need test for this (`#` literal).

tools/p4mlir-translate/translate.cpp
```diff
@@ -1139,21 +1224,20 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
...
-            } else if (bCall->name == P4::IR::Type_Header::isValid) {
-                auto validityBit = b.create<P4HIR::ReadOp>(loc, validityBitRef);
-                callResult = builder.create<P4HIR::CmpOp>(loc, P4HIR::CmpOpKind::Eq, validityBit,
-                                                          getValidHeaderConstant(loc));
-            } else
-                BUG("Unsupported builtin method: %1%", bCall->name);
+
+            // Check if this is a method call on a header or header union
+            if (mce->method->is<P4::IR::Member>()) {
+                const auto *member = mce->method->to<P4::IR::Member>();
```
asl, line 1230: you can just do `if (const auto *member = mce->method->to<P4::IR::Member>()) { ... }`

tools/p4mlir-translate/translate.cpp
```diff
@@ -1139,21 +1224,20 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
...
+                const auto *exprType = typeMap->getType(member->expr);
```
asl, line 1231: Do not use typemap unless necessary. Is `member->expr->type` invalid?

tools/p4mlir-translate/translate.cpp
```diff
@@ -1139,21 +1224,20 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
...
+
+                // Check if it's a reference to a header or header union
+                if (const auto *const p4cHeaderUnionType =
+                        exprType->to<P4::IR::Type_HeaderUnion>()) {
+                    callResult = emitHeaderUnionBuiltInMethod(p4cHeaderUnionType, bCall, b, loc);
+                } else if (const auto *p4cHeaderType = exprType->to<P4::IR::Type_Header>()) {
```
asl, line 1237: `p4cHeaderType` is unused

tools/p4mlir-translate/translate.cpp
```diff
@@ -1276,7 +1360,13 @@ void P4HIRConverter::postorder(const P4::IR::Member *m) {
     // TODO: Likely we can do similar things for the majority of struct-like
     // types
     auto parentType = getOrCreateType(m->expr);
-    if (mlir::isa<P4HIR::StructType, P4HIR::HeaderType>(parentType)) {
+    if (mlir::isa<P4HIR::StructType, P4HIR::HeaderUnionType>(parentType)) {
```
asl, line 1363: why duplicate the code? Just add another case to `isa`

tools/p4mlir-translate/translate.cpp
```diff
@@ -1139,21 +1224,20 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
             // TODO: Are there cases when we do not have l-value here?
             auto ref = resolveReference(bCall->appliedTo);
```
asl, line 1225: `ref` is unused?

tools/p4mlir-translate/translate.cpp
```diff
@@ -1115,6 +1143,63 @@ void P4HIRConverter::postorder(const P4::IR::ReturnStatement *ret) {
...
+    } else {
+        BUG("Unsupported builtin method: %1%", builtInMethod->name);
+    }
+
+    return callResult;
+}
+
+mlir::Value P4HIRConverter::emitHeaderUnionBuiltInMethod(
+    const P4::IR::Type_HeaderUnion *p4cHeaderUnionType, const P4::BuiltInMethod *builtInMethod,
+    mlir::OpBuilder &builder, mlir::Location loc) {
```
asl, line 1172: no need to pass `builder `here (and above).

tools/p4mlir-translate/translate.cpp
```diff
@@ -1115,6 +1143,63 @@ void P4HIRConverter::postorder(const P4::IR::ReturnStatement *ret) {
...
+
+    auto isHeaderUnionValid = getBoolConstant(loc, false);  // Start with a false value
+    auto headerUnion = resolveReference(builtInMethod->appliedTo);
+    auto dialectHeaderUnionType =
+        mlir::cast<P4HIR::HeaderUnionType>(getOrCreateType(p4cHeaderUnionType));
+    auto huFields = dialectHeaderUnionType.getFields();
+
+    for (auto huField : huFields) {
+        auto headerRef = builder.create<P4HIR::StructExtractRefOp>(loc, headerUnion, huField.name);
+        auto validityBitRef = builder.create<P4HIR::StructExtractRefOp>(
```
asl, line 1184: Maybe you can refactor the code a bit, so you could reuse some pieces of `emitHeaderBuiltInMethod`?

E.g.:
 - Instead of passing `builtInMethod`, pass only method name
 - Instead of emitting a reference to a header (via `resolveReference`), inside do it outside, and pass the `mlir::Value` as argument.

This way you'd essentially need to call such refactored `emitHeaderBuiltInMethod` with `headerRef` argument and then combine the resulting values properly.

tools/p4mlir-translate/translate.cpp
```diff
@@ -1115,6 +1143,63 @@ void P4HIRConverter::postorder(const P4::IR::ReturnStatement *ret) {
...
+            loc, headerRef, P4HIR::HeaderType::validityBit);
+        auto validityBit = builder.create<P4HIR::ReadOp>(loc, validityBitRef);
+        auto isHeaderValid = builder.create<P4HIR::CmpOp>(loc, P4HIR::CmpOpKind::Eq, validityBit,
+                                                          getValidHeaderConstant(loc));
+        auto ternaryOpResult = builder.create<P4HIR::TernaryOp>(
```
asl, line 1189: See above about expected code sequence.

asl (changes requested):

test/Translate/Ops/header_union.p4
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done

include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
```diff
@@ -302,10 +302,17 @@ class StructLikeType<string name, string typeMnemonic>
        return {};
     }
 
-    llvm::ArrayRef<P4HIR::FieldInfo> getFields() {
+    llvm::ArrayRef<P4HIR::FieldInfo> getFields() const {
       return getElements();
     }
 
+    template <typename Func>
+    void forEachField(Func &&func) const {
```
asl, line 310: Do we really need this function? Can we use `llvm::for_each(getFields(), func)` instead?

lib/Dialect/P4HIR/P4HIR_Types.cpp
```diff
@@ -299,8 +306,37 @@ LogicalResult HeaderType::verify(function_ref<InFlightDiagnostic()> emitError, S
...
+                        << "' must be a header type";
+        }
+    }
+
+    if (elements.empty()) {
+        emitError() << "empty p4hir.header_union type";
+        return failure();
+    }
+
+    return result;
```
asl, line 334: You also need to check that each field is of `HeaderType`

tools/p4mlir-translate/translate.cpp
```diff
@@ -189,9 +190,99 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+        auto validityBitConstant = emitValidityConstant(loc, validityConstValue);
+        auto validityBitRef =
+            builder.create<P4HIR::StructExtractRefOp>(loc, header, P4HIR::HeaderType::validityBit);
+        builder.create<P4HIR::AssignOp>(loc, validityBitConstant, validityBitRef);
+    }
+
+    P4HIR::CmpOp emitHeaderIsValidCmpOp(mlir::Location loc, mlir::Value header,
+                                        P4HIR::ValidityBit validityConstValue, bool useRefOp) {
+        mlir::Value validityBitValue;
+        if (useRefOp) {
```
asl, line 209: Can't you deduce this from the type of `header`? If this is of `ReferenceType`, then you'd need to use `StructExtractRefOp` otherwise – `StructExtractOp`.

tools/p4mlir-translate/translate.cpp
```diff
@@ -189,9 +190,99 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+                builder.create<P4HIR::StructExtractOp>(loc, header, P4HIR::HeaderType::validityBit);
+        }
+        auto validityConstant = emitValidityConstant(loc, validityConstValue);
+        return builder.create<P4HIR::CmpOp>(loc, P4HIR::CmpOpKind::Eq, validityBitValue,
+                                            validityConstant);
+    }
+
+    P4HIR::CmpOp emitHeaderUnionIsValidCmpOp(mlir::Location loc, mlir::Value headerUnion,
+                                             P4HIR::HeaderUnionType headerUnionType,
+                                             P4HIR::ValidityBit validityBitValue, bool useRefOp) {
```
asl, line 224: See above for `useRefOp`

tools/p4mlir-translate/translate.cpp
```diff
@@ -449,13 +544,21 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
     bool preorder(const P4::IR::Member *m) override;
     bool preorder(const P4::IR::Equ *) override;
     bool preorder(const P4::IR::Neq *) override;
+    bool preorderInvalidHeader(const P4::IR::Operation_Relation *p4RelationOp);
```
asl, line 547: I'd not mix overridden methods with normal ones.

tools/p4mlir-translate/translate.cpp
```diff
@@ -993,41 +1117,43 @@ CONVERT_CMP(Geq, Ge);
...
-            eq, builder.create<P4HIR::CmpOp>(loc, P4HIR::CmpOpKind::Eq, validityBitField, invalid));
-        return false;
-    }
-
-    return true;
+bool P4HIRConverter::preorderInvalidHeader(const P4::IR::Operation_Relation *p4RelationOp) {
+    auto loc = getLoc(builder, p4RelationOp);
+    visit(p4RelationOp->left);
+    auto validityConstValue =
+        p4RelationOp->to<P4::IR::Equ>() ? P4HIR::ValidityBit::Invalid : P4HIR::ValidityBit::Valid;
```
asl, line 1124:
```suggestion
        p4RelationOp->is<P4::IR::Equ>() ? P4HIR::ValidityBit::Invalid : P4HIR::ValidityBit::Valid;
```

tools/p4mlir-translate/translate.cpp
```diff
@@ -189,9 +190,99 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+
+    void emitAllHeadersInvalidForParentHeaderUnion(mlir::Location loc, const P4::IR::Expression *header) {
+        if (const auto *member = header->to<P4::IR::Member>()) {
+            if (member->expr != nullptr && member->expr->type != nullptr) {
+                if (const auto *p4cHeaderUnionType =
+                        member->expr->type->to<P4::IR::Type_HeaderUnion>()) {
+                    auto headerUnionType =
+                        mlir::cast<P4HIR::HeaderUnionType>(getOrCreateType(p4cHeaderUnionType));
+                    auto headerUnion = resolveReference(member->expr);
+                    headerUnionType.forEachField([&](P4HIR::FieldInfo fieldInfo) {
```
asl, line 278: See above about `forEachField`

tools/p4mlir-translate/translate.cpp
```diff
@@ -189,9 +190,99 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+        auto isValid = buildNestedTernaryOp(0);
+
+        // Return a comparison operation for consistency with other validity checks
+        return builder.create<P4HIR::CmpOp>(loc, P4HIR::CmpOpKind::Eq, isValid,
+                                            getBoolConstant(loc, true));
+    }
+
+    void emitAllHeadersInvalidForParentHeaderUnion(mlir::Location loc, const P4::IR::Expression *header) {
+        if (const auto *member = header->to<P4::IR::Member>()) {
+            if (member->expr != nullptr && member->expr->type != nullptr) {
```
asl, line 272: This should never happen. In which case you're seeing `IR::Member` with null parent?

tools/p4mlir-translate/translate.cpp
```diff
@@ -189,9 +190,99 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+                if (const auto *p4cHeaderUnionType =
+                        member->expr->type->to<P4::IR::Type_HeaderUnion>()) {
+                    auto headerUnionType =
```
asl, line 275: So, why can't you simply grab type out of `headerUnion` below? E.g.:

```c++
if (member->expr->type->is<P4::IR::Type_HeaderUnion>()) {
  auto headerUnion = resolveReference(member->expr);
  auto huType = mlir::cast<P4HIR::HeaderUnionType>(headerUnion.getType());
  llvm:for_each(huType.getFields(), [&]() { ... });
```

tools/p4mlir-translate/translate.cpp
```diff
@@ -1036,13 +1162,22 @@ bool P4HIRConverter::preorder(const P4::IR::AssignmentStatement *assign) {
...
-        builder.create<P4HIR::AssignOp>(loc, invalid, validityBitRef);
+        // Assignment of InvalidHeader is special: we do not materialize the whole
+        // header, we need to assign validty bit only
+        emitHeaderValidityBitAssignOp(loc, ref, P4HIR::ValidityBit::Invalid);
+    } else if (const auto *const p4cHeaderUnionType =
+                   assign->right->type->to<P4::IR::Type_HeaderUnion>()) {
+        // Assignment of InvalidHeaderUnion is special: all headers in the union will be invalidated
+        auto dialectHeaderUnionType =
+            mlir::cast<P4HIR::HeaderUnionType>(getOrCreateType(p4cHeaderUnionType));
+        dialectHeaderUnionType.forEachField([&](P4HIR::FieldInfo fieldInfo) {
```
asl, line 1177: See above, you do not need to materialize type, get it from the value, it is already there!

tools/p4mlir-translate/translate.cpp
```diff
@@ -1326,6 +1461,41 @@ void P4HIRConverter::postorder(const P4::IR::ReturnStatement *ret) {
...
+    return callResult;
+}
+
+mlir::Value P4HIRConverter::emitHeaderUnionBuiltInMethod(
+    mlir::Location loc, const P4::BuiltInMethod *builtInMethod,
+    const P4::IR::Type_HeaderUnion *p4cHeaderUnionType) {
+    if (builtInMethod->name != P4::IR::Type_Header::isValid)
+        BUG("Unsupported Header Union builtin method: %1%", builtInMethod->name);
+    auto headerUnion = resolveReference(builtInMethod->appliedTo);
+    auto headerUnionType = mlir::cast<P4HIR::HeaderUnionType>(getOrCreateType(p4cHeaderUnionType));
```
asl, line 1494: See above about type. And you do not need to pass it from `emitHeaderUnionIsValidCmpOp` either.

asl (commented):
See comments

include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
tagolog: The function removed.

lib/Dialect/P4HIR/P4HIR_Types.cpp
tagolog: There is a check for the header type:
https://github.com/tagolog/p4mlir-incubator/blob/mlir-header-union/lib/Dialect/P4HIR/P4HIR_Types.cpp#L321-L326

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Refactored

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done. I've implemented the template method "getType" to handle the "ReferenceType" in a proper manner.

tools/p4mlir-translate/translate.cpp
tagolog: Done. See the comment above.

tools/p4mlir-translate/translate.cpp
tagolog: Done. See the comment above.

include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
```diff
@@ -302,7 +302,7 @@ class StructLikeType<string name, string typeMnemonic>
        return {};
     }
 
-    llvm::ArrayRef<P4HIR::FieldInfo> getFields() {
+    llvm::ArrayRef<P4HIR::FieldInfo> getFields() const {
```
asl, line 305: no need for this

tagolog: Done

tools/p4mlir-translate/translate.cpp
```diff
@@ -3,6 +3,8 @@
 #include <algorithm>
 #include <climits>
 
+#include "ir/ir-generated.h"
```
asl, line 6: Please remove

tools/p4mlir-translate/translate.cpp
```diff
@@ -467,6 +558,20 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
     mlir::Value emitCmp(const P4::IR::Operation_Relation *relop, P4HIR::CmpOpKind kind);
+
+ private:
+    mlir::Value emitInvalidHeaderCmpOp(const P4::IR::Operation_Relation *p4RelationOp);
+    mlir::Value emitInvalidHeaderUnionCmpOp(const P4::IR::Operation_Relation *p4RelationOp);
+    mlir::Value emitHeaderBuiltInMethod(mlir::Location loc, const P4::BuiltInMethod *builtInMethod);
+    mlir::Value emitHeaderUnionBuiltInMethod(mlir::Location loc,
+                                             const P4::BuiltInMethod *builtInMethod);
+    template <typename DialectType>
+    DialectType getType(mlir::Value &value) {
```
asl, line 569: I would name it differently, as you're getting type via `value.getType()`, so here the name is misleading. Maybe `getObjectType()`?

tools/p4mlir-translate/translate.cpp
```diff
@@ -467,6 +558,20 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+        return mlir::isa<P4HIR::ReferenceType>(value.getType())
```
asl, line 570:
```suggestion
        if (auto refType = mlir::dyn_cast<P4HIR::ReferenceType>(value.getType()))
           return refType.getObjectType();
        return value.getType();
```

The explicit cast is quite misleading here IMO. I would just make `getObjectType` return `mlir::Type` and perform cast at caller side.

tools/p4mlir-translate/translate.cpp
```diff
@@ -1032,57 +1149,80 @@ CONVERT_CMP(Geq, Ge);
...
 
-    return true;
+    visit(relOp->left);
+
+    if (relOp->is<P4::IR::Equ>()) {
+        return emitHeaderIsValidCmpOp(loc, header, P4HIR::ValidityBit::Invalid);
+    } else if (relOp->is<P4::IR::Neq>()) {
+        return emitHeaderIsValidCmpOp(loc, header, P4HIR::ValidityBit::Valid);
+    }
+    BUG("unexpected relation operator %1%", relOp->node_type_name());
```
asl, line 1163:
```suggestion
    BUG("unexpected relation operator %1%", relOp);
```

tools/p4mlir-translate/translate.cpp
```diff
@@ -1032,57 +1149,80 @@ CONVERT_CMP(Geq, Ge);
...
-    }
+    visit(relOp->left);
 
-    return true;
+    if (relOp->is<P4::IR::Equ>()) {
+        return emitHeaderUnionIsValidCmpOp(loc, headerUnion, P4HIR::ValidityBit::Invalid);
+    } else if (relOp->is<P4::IR::Neq>()) {
+        return emitHeaderUnionIsValidCmpOp(loc, headerUnion, P4HIR::ValidityBit::Valid);
+    }
+    BUG("unexpected relation operator %1%", relOp->node_type_name());
```
asl, line 1177:
```suggestion
    BUG("unexpected relation operator %1%", relOp);
```

tools/p4mlir-translate/translate.cpp
```diff
@@ -1365,6 +1505,43 @@ void P4HIRConverter::postorder(const P4::IR::ReturnStatement *ret) {
...
+mlir::Value P4HIRConverter::emitHeaderBuiltInMethod(mlir::Location loc,
+                                                    const P4::BuiltInMethod *builtInMethod) {
+    mlir::Value callResult;
+    auto header = resolveReference(builtInMethod->appliedTo);
+    if (builtInMethod->name == P4::IR::Type_Header::setValid ||
+        builtInMethod->name == P4::IR::Type_Header::setInvalid) {
+        // Check if the header is a member of a header union
+        if (const auto *member = builtInMethod->appliedTo->to<P4::IR::Member>()) {
+            if (member->expr->type->is<P4::IR::Type_HeaderUnion>()) {
+                const auto headerNameToSkip = builtInMethod->name == P4::IR::Type_Header::setValid
```
asl, line 1517: `member->member.name` is a `cstring`. Consider using it directly instead of allocating and copying to std::string. You can use `nullptr` instead of `""` here. And you can use `cstring::isNullOrEmpty` to check if there is something in the string inside `emitSetInvalidForAllHeaders`.

Since `cstring` is essentially a pointer, it would be passed into `emitSetInvalidForAllHeaders` by value.

tools/p4mlir-translate/translate.cpp
```diff
@@ -1471,6 +1638,7 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
             operands.push_back(argVal);
         }
 
+        mlir::Value callResult;
```
asl, line 1641: unused?

tools/p4mlir-translate/translate.cpp
```diff
@@ -192,9 +195,93 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+        auto isValid = buildNestedTernaryOp(0);
+
+        // Return a comparison operation for consistency with other validity checks
+        return builder.create<P4HIR::CmpOp>(
+            loc, P4HIR::CmpOpKind::Eq, isValid,
+            getBoolConstant(loc, compareWith == P4HIR::ValidityBit::Valid ? true : false));
+    }
+
+    void emitSetInvalidForAllHeaders(mlir::Location loc, mlir::Value headerUnion,
+                                     const std::string headerNameToSkip = "") {
```
asl, line 276: use `cstring` here.

tools/p4mlir-translate/translate.cpp
```diff
@@ -192,9 +195,93 @@ class P4HIRConverter : public P4::Inspector, public P4::ResolutionContext {
...
+        auto headerUnionType = getType<P4HIR::HeaderUnionType>(headerUnion);
+        llvm::for_each(headerUnionType.getFields(), [&](P4HIR::FieldInfo fieldInfo) {
+            if (headerNameToSkip.size() == 0 || fieldInfo.name != headerNameToSkip) {
```
asl, line 279:
```suggestion
            if (headerNameToSkip != fieldInfo.name) {
```

(for `cstring headerNameToSkip`).

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Used below. We can't remove it.

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done

tools/p4mlir-translate/translate.cpp
tagolog: Done

asl (approved):
Yay!

PR #102, opened 2025-03-07
Fix MacOS m chip build script

Joejiong:
Minor change to the build script, and solve the issue with https://github.com/p4lang/p4c/issues/5166

if somebody is going to build p4mlir, they need to put the necessary code into or organize it in a separate place, so they would not conflict with homebrew-installed ones: 

```
$ which bison
/opt/local/bin/bison
$ which flex
/opt/local/bin/flex
```
macOS building successfully:
```
3/4] Running the P4MLIR regression tests

Testing Time: 0.93s

Total Discovered Tests: 45
  Passed: 45 (100.00%)
```

### DIFF
--- a/build_tools/build_p4c_with_p4mlir_ext.sh
+++ b/build_tools/build_p4c_with_p4mlir_ext.sh
@@ -20,14 +20,8 @@ P4C_EXT_DIR=$P4C_REPO_DIR/extensions
 
 P4C_P4MLIR_EXT_DIR=$P4C_EXT_DIR/p4mlir
 
-# if [[ "$(uname)" == "Darwin" ]]; then
-#     # macOS version
 P4C_P4MLIR_EXT_SYMLINK=$(realpath --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR" 2>/dev/null || \
 python3 -c 'import os, sys; print(os.path.relpath(sys.argv[2], start=sys.argv[1]))' "$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
-# else
-#     # Linux or other OS
-#     P4C_P4MLIR_EXT_SYMLINK=$(realpath -s --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
-# fi
 
 # Link P4MLIR as P4C extension
 mkdir -p "$P4C_EXT_DIR"
--- a/build_tools/build_p4c_with_p4mlir_ext.sh
+++ b/build_tools/build_p4c_with_p4mlir_ext.sh
@@ -19,7 +19,15 @@ P4C_BUILD_DIR=$P4C_REPO_DIR/build
 P4C_EXT_DIR=$P4C_REPO_DIR/extensions
 
 P4C_P4MLIR_EXT_DIR=$P4C_EXT_DIR/p4mlir
-P4C_P4MLIR_EXT_SYMLINK=$(realpath -s --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
+
+if [[ "$(uname)" == "Darwin" ]]; then
+    # macOS version
+    P4C_P4MLIR_EXT_SYMLINK=$(realpath --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR" 2>/dev/null || \
+    python3 -c 'import os, sys; print(os.path.relpath(sys.argv[2], start=sys.argv[1]))' "$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
+else
+    # Linux or other OS
+    P4C_P4MLIR_EXT_SYMLINK=$(realpath -s --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
+fi
 
 # Link P4MLIR as P4C extension
 mkdir -p "$P4C_EXT_DIR"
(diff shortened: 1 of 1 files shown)


build_tools/build_p4c_with_p4mlir_ext.sh
```diff
@@ -19,7 +19,15 @@ P4C_BUILD_DIR=$P4C_REPO_DIR/build
 P4C_EXT_DIR=$P4C_REPO_DIR/extensions
 
 P4C_P4MLIR_EXT_DIR=$P4C_EXT_DIR/p4mlir
-P4C_P4MLIR_EXT_SYMLINK=$(realpath -s --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
+
+if [[ "$(uname)" == "Darwin" ]]; then
+    # macOS version
+    P4C_P4MLIR_EXT_SYMLINK=$(realpath --relative-to="$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR" 2>/dev/null || \
+    python3 -c 'import os, sys; print(os.path.relpath(sys.argv[2], start=sys.argv[1]))' "$(dirname "$P4C_P4MLIR_EXT_DIR")" "$P4MLIR_REPO_DIR")
```
qobilidop, line 26: I did some search on this and learned that the root cause is there are 2 different `realpath` commands. On Linux, [`realpath(1)`](https://man7.org/linux/man-pages/man1/realpath.1.html) (the GNU version) is the default while on macOS [`realpath(3)`](https://man7.org/linux/man-pages/man3/realpath.3.html) (the BSD version) is the default. They accept different options and arguments so the original shell command is not portable.

There are probably 2 other solutions which are cleaner:
1. We could add some notes on macOS installation that `brew install coreutils` is required. This could bring `realpath(1)` to macOS and the original shell command shall work.
2. If the Python solution is portable, we could use it on both macOS and Linux cases. This could simplify the script.

@Joejiong What do you think?

qobilidop:
Not a request to add them in this PR, but in general, if we'd like to support macOS project building, it would be nice to have:
1. A macOS build guide.
2. A GitHub Actions workflow that tests building the project on macOS.

build_tools/build_p4c_with_p4mlir_ext.sh
asl: Frankly speaking, these scripts are already pretty confusing and error prone. E.g. as project is linked inside `p4c` tree you'd need to edit project files also inside the `p4c` tree as all source paths are all relative to `p4c` tree. This includes compilation databases, binaries, etc.

I would probably suggest to go other way and to try to get rid of the script entirely trying to organize everything is a bit different way. Maybe we can do some magic with git submodules?

qobilidop: > 1. We could add some notes on macOS installation that `brew install coreutils` is required. This could bring `realpath(1)` to macOS and the original shell command shall work.

Just realized what's installed this way is called `grealpath`. So the original command is still not portable. Then maybe 2 is a better solution.

qobilidop: > Frankly speaking, these scripts are already pretty confusing and error prone. E.g. as project is linked inside `p4c` tree you'd need to edit project files also inside the `p4c` tree as all source paths are all relative to `p4c` tree. This includes compilation databases, binaries, etc.
> 
> I would probably suggest to go other way and to try to get rid of the script entirely trying to organize everything is a bit different way. Maybe we can do some magic with git submodules?

Yeah, it's worth discussing how things could be better organized in the longer term. But probably not for this PR though.

I don't quite like how the P4C extension system requires putting code under the `p4c/extensions/` directory. I prefer the LLVM/MLIR out-of-tree extension approach more. They don't have such directory requirement. If possible, I'd like to change the P4C side to allow developing extensions in arbitrary directories.

Joejiong:
> Not a request to add them to this PR, but in general, if we'd like to support macOS project building, it would be nice to have:
> 
> 1. A macOS build guide.
> 2. A GitHub Actions workflow that tests building the project on macOS.

OK, I will leave it as a future reference.

build_tools/build_p4c_with_p4mlir_ext.sh
Joejiong: Sure I will leave this for the future.

> > Frankly speaking, these scripts are already pretty confusing and error prone. E.g. as project is linked inside `p4c` tree you'd need to edit project files also inside the `p4c` tree as all source paths are all relative to `p4c` tree. This includes compilation databases, binaries, etc.
> > I would probably suggest to go other way and to try to get rid of the script entirely trying to organize everything is a bit different way. Maybe we can do some magic with git submodules?
> 
> Yeah, it's worth discussing how things could be better organized in the longer term. But probably not for this PR though.
> 
> I don't quite like how the P4C extension system requires putting code under the `p4c/extensions/` directory. I prefer the LLVM/MLIR out-of-tree extension approach more. They don't have such directory requirement. If possible, I'd like to change the P4C side to allow developing extensions in arbitrary directories.

fruffy: > I don't quite like how the P4C extension system requires putting code under the `p4c/extensions/` directory. I prefer the LLVM/MLIR out-of-tree extension approach more. They don't have such directory requirement. If possible, I'd like to change the P4C side to allow developing extensions in arbitrary directories.

This could be as easy as adding an option for a new extension folder. Or adding a list of folders that are extensions. Although symlinking has worked well so far. 

The  one snag is to make sure that the linters/analysis tools still work correctly.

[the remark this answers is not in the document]
Joejiong: OK, let me try, and test for the Linux machine.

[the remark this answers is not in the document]
Joejiong: > @Joejiong There might be too many discussions above. The concrete suggestion I have for this PR is, could you try using this Python command alone for both macOS and Linux cases? I feel there is no need to have 2 conditions here.

@qobilidop it succeeded.

qobilidop (approved):

PR #107, opened 2025-03-10
typevar check v1

Joejiong:
This is just a placeholder for discussion. I have done some research about the p4 extern function context and with your newly added support for extern func support at p4milr. 

Now the check will happen in three different places:
1. CallMethodOp
2. ExternOp
3. FuncOp

The current TODO verification in typevartype might lack context to solve the issue everywhere.
```
// CHECK:   p4hir.func @externFunc<!p4hir.type_var<"T">, !p4hir.type_var<"U">>(!p4hir.type_var<"U"> {p4hir.dir = #in}, !p4hir.type_var<"T"> {p4hir.dir = #in}) -> !p4hir.type_var<"T">
extern T externFunc<T, U>(in U a, in T b);
current output:

Failed Tests (4):
  P4MLIR :: Dialect/P4HIR/extern.mlir
  P4MLIR :: Translate/Ops/extern.p4
  P4MLIR :: Translate/Ops/extern_func.p4
  P4MLIR :: Translate/Parser/packet_in.p4


Testing Time: 2.22s

Total Discovered Tests: 52
  Passed: 48 (92.31%)
  Failed:  4 (7.69%)
```

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
@@ -1125,7 +1125,7 @@ def ExternOp : P4HIR_Op<"extern",
   let assemblyFormat = [{
     $sym_name (`<` $type_parameters^ `>`)? $body attr-dict
   }];
-
+  let hasVerifier = 1;
   let extraClassDeclaration = [{
     mlir::Block &createEntryBlock();
   }];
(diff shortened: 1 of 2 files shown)


qobilidop:
Hi @Joejiong, how's it going? Are you able to make the CI pass as a first step?

Joejiong:
I'm still working on how to chunk the task, cause there are two ways to do it just by imagination, 1. Check from call sites or 2. from type itself. these verifications will change some tests in these files, I probably need to book a meeting to discuss whether my understandings are correct for the newly added externOp and packet_in support. I have another issue which is what are some checks criteria we need to put in, for example, when Anton says:
1. Ensure that type variables are only used on declarations
2. Check method destinations.

The current code is for typed variables not in return, am I right for this and I will gradually add other checks and also modify their tests, I'm also trying to understand the broader picture of p4 extern func use cases. @qobilidop

asl:
> these verifications will change some tests in these files, I

All current tests are correct and should not cause any verification fails. If this happens then it is a clear indicator that something is not correct.

PR #134, opened 2025-03-21
Add support for constant annotations

asl:


test/Translate/Ops/annotations.p4
```diff
@@ -0,0 +1,107 @@
...
+// CHECK:   p4hir.const ["b"] #int-1_b1i annotations {hidden}
+@hidden const bit b = 1;
+
+struct Foo {
+  @match(ternary) bit<32> bar;
+}
+
+// CHECK-LABEL  p4hir.extern @Annotated annotations {size = ["100"]} {
+// CHECK:    p4hir.func @Annotated() annotations {hidden, name = "annotated", pkginfo = {bar = "42", foo = 10 : i64}}
+// CHECK:    p4hir.func @execute(!b8i {p4hir.annotations = {optional}, p4hir.dir = #undir, p4hir.param_name = "index"}) annotations {name = "exe"}
```
Joejiong, line 20: so you have chosen that optional will be a mlir builtin optionalattr, then the actual handling will wait until the actual backend?

asl: It is not a builtin. We translate and preserve all annotations even those that should be handled in the frontend (e.g. `@tableonly` or `@defaultonly`). We still need to handle optional argument semantics.

qobilidop (approved):

PR #141, opened 2025-03-26
Ensure we correctly unify IntAttr with infint type

asl:


qobilidop:
Would like to have a better understanding of the problem. When would `APInt::operator==` be used in the code?

> The problem is that assertion might be non-deterministic as `operator==` is only called on hash collisions

I don't follow this part. I guess it's related to my question above. Maybe an example could help me here.

asl:
> Would like to have a better understanding of the problem. When would `APInt::operator==` be used in the code?

Attributes are unique. So there will be a value unification code produced by tablegen. However, default unification code is not suitable for us as it will essentially be:
```c++
    bool operator==(const KeyTy &rhs) const {
        auto rhsType = std::get<0>(rhs);
        const auto &rhsValue = std::get<1>(rhs);
        return (type == rhsType) && (value == rhsValue);
    }
```

Note that `value == rhsValue` is unguarded here, so values of different bitwidths could be compared here.

These values are stored in a large `DenseMap`. And the hashes of the fields are combined via LLVM's `hash_combine`. The latter is non-deterministic (by purpose), it includes random seed into the hash. So, the `operator==` is only called on hash collisions. So the assertion is a bit non-deterministic.

qobilidop:
Thanks for the detailed explanation!

qobilidop (approved):

Issue #136, opened 2025-03-21
Decide on set of frontend passes to run before dialect translation


asl:
Currently frontend pipeline contains lots of intermixed passes that do very different things:
 - Perform type inference
 - Do some diagnostics
 - Perform some desugaring / canonicalization
 - Perform some optimizations / code folding

Some of passes do many things at the same time. Some known to be error prone, have suboptimal runtime and memory consumption, some definitely do not belong to the frontend. We need to decide what we'd need to run before dialect translation. Overall, the less we're running the better as P4C passes have non-negligible overhead as every `Transform` effectively clone the whole program (even w/o any changes). So we'd need to keep the things that are necessary for sane translation and if there is something that could be done in MLIR as a dialect transform, it would be better to do it there. Overall, certainly we'd run every `Inspector` that validates something.

Here is the breakdown with my notes:

  * ParseAnnotationBodies – *definitely need it*
  * ValidateParsedProgram - use it
  * CreateBuiltins - use it
  * CheckShadowing - a bit expensive, but required for diagnostics
  * ConstantFolding –  First pass of constant folding, before types are known, necessary for type inference
  * ValidateStringAnnotations – Validate `@name/@deprecated/@noWarn`. Should run after constant folding.
  * InstantiateDirectCalls - simplifies control lowering, use it
  * Deprecated - use it
  * CheckNamedArgs - use it
  * TypeInference - use it
  * ValidateMatchAnnotations - use it
  * ValidateValueSets - use it
  * DefaultValues - use it
  * BindTypeVariables - use it
  * EntryPriorities - use it
  * Generics specialization - use it
  * CheckCoreMethods - use it
  * StaticAssert - use it
  * RemoveParserIfs - looks like a bit too early? Better be done at MLIR level #99 
  * StructInitializers - use it, simplifies some initializers translation
  * TableKeyNames - use it
  * PassRepeated: - no need, do MLIR folding
    * ConstantFolding
    * StrengthReduction
    * Reassociation
    * UselessCasts
  * SimplifyControlFlow - no need, do MLIR folding
  * SwitchAddDefault - could be done at MLIR level
  * RemoveAllUnusedDeclarations - could be done at MLIR level
  * SimplifyParsers - could be done at MLIR level
  * ResetHeaders - could be done during translation, #116
  * UniqueNames - no need
  * MoveDeclarations - no need
  * MoveInitializers - no need, if necessary, move to MLIR level.
[...]

PR #142, opened 2025-03-26
Add some set of frontend passes

asl:
For https://github.com/p4lang/p4mlir-incubator/issues/136


qobilidop (approved):
If @fruffy could take a look, that would be better.

asl:
@fruffy will you please take a look?

Issue #151, opened 2025-04-01
Set up starter code for the BMv2 code exporter


qobilidop:
This is the beginner step of #139.

`ExportSystemC` from [CIRCT](https://circt.llvm.org/) is a good example on how to structure the code. Search for `--export-systemc` and `ExportSystemC` in https://github.com/llvm/circt to draw some inspirations.

For this task, the goal is the make the exporter work with a very simple P4 program. We don't need a BMv2 dialect yet. Just directly translate from P4HIR to BMv2 JSON. This P4 program could be as simple as a single global `const` definition. Try using the P4C BMv2 backend to get the expected output JSON first to guide the development.

Some related MLIR glossory:
- [Export](https://mlir.llvm.org/getting_started/Glossary/#export)
- [Translation](https://mlir.llvm.org/getting_started/Glossary/#translation)

qobilidop:
@asl Do you have a preference if we should directly register the exporter in `p4mlir-translate`? Or prototype in a separate binary first like `p4mlir-export-bmv2` maybe?

I slightly prefer starting with a separate binary and merge it into `p4mlir-translate` later if you don't have major concerns.

asl:
Yeah, I think it should be separate. Just load P4HIR (from bytecode / .mlir source) and produce whatever is necessary

qobilidop:
@Joejiong Do you want to take a look? This could qualify you for [project 7](https://github.com/p4lang/gsoc/blob/main/2025/ideas_list.md#project-7) (but not project 6, which requires your contribution to the core P4HIR, like `@optional`).

Joejiong:
> [@Joejiong](https://github.com/Joejiong) Do you want to take a look? This could qualify you for [project 7](https://github.com/p4lang/gsoc/blob/main/2025/ideas_list.md#project-7) (but not project 6, which requires your contribution to the core P4HIR, like `@optional`).
[WIP] Collecting ideas.
This is not an RFC, still collecting ideas:
Yes, for sure, previously, I thought the bmv2 dialect would be designed as closely as the JSON file, which could be used for simple_switch, cause the backend is close to running model, it has a more static pipeline, and we could have a general Dialect2JsonConvertor, all others could extend this to their own. 

The following could work for bmv2.json like dialects, but for the frontend, it is a bit difficult to integrate all together. directly transfer from p4hir would be ok, but somehow feels too early in the pipeline, but p4hir could also extend the following unified design to generate its json.

[...]

PR #154, opened 2025-04-07
[WIP] Export tool starter, trial and error code

Joejiong:
[https://github.com/p4lang/p4mlir-incubator/issues/151]
I have another experimental code in progress not yet uploaded, and it uses another approach.


tools/p4mlir-export/simplebmv2.p4
```diff
@@ -0,0 +1,69 @@
+// Simplest p4 program to be processed, which can go through
```
fruffy, line 1: Why is the mlir code for this file empty? Does it not compile? What is the expected json output for it? Maybe you should add it as a reference.

Joejiong: oh, forgot to upload. let me upload it

fruffy: Try this code as example: https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/bmv2_extract_1.p4

I know this already compiles to mlir and it is a very simple example of a BMV2 program. Then compile this program and generate the json file using p4c-bm2-ss. You should be able to generate basic elements already. 

I am not sure what looped references or typedefs are.

fruffy: How to produce text output from this particular MLIR dialect you can check https://github.com/p4lang/p4mlir-incubator/blob/fruffy/top4/lib/Utilities/mlir_to_p4.h#L20. It writes MLIR to P4 again. This code is WiP, by no means complete, and will change. But it is a start and shows that it can be done.

Joejiong: > I am not sure what looped references or typedefs are.
For example:
```
!b16i = !p4hir.bit<16>
```
After the p4mlir-translate, the code gets expanded, and also linked with other v1 arch codes.
for loop reference, I mean, it is dealing with my traverse methods, it is not a tree, could have nodes being traversed.
My current code hangs at complex p4hir, goes into an infinite loop. Lastly, the loop syntax they currently support would not be easily represented as a JSON DAG if it is not deterministic.

Sure, thx, I will check your code.

fruffy: Don't worry about for loops for now, there is very little Bmv2 code written that way. Just throw an error that it is not implemented,

Joejiong: > Don't worry about for loops for now, there is very little Bmv2 code written that way. Just throw an error that it is not implemented,

got you, thx

tools/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 add_subdirectory(p4mlir-translate)
 add_subdirectory(p4mlir-opt)
+add_subdirectory(p4mlir-export)
```
qobilidop, line 3: Could we rename this to `p4mlir-export-bmv2` to be more specific?

.gitignore
```diff
@@ -4,4 +4,5 @@ install/
 .vscode/
 .cursor/
 .idea/
+.DS_Store
```
qobilidop, line 7: I feel it's better to ignore this in your system-wide `.gitnigore` instead of here.

tools/p4mlir-export/README
qobilidop, line 1: Could you rename this file to `README.md` to be more consistent with the rest of the repo?

tools/p4mlir-export/bool_test.json
qobilidop, line 1: I don't feel your `p4hir-iso-json` approach is necessary. The `.mlir` format is already a standard textual format, and is more readable than this JSON representation.

We can have further discussions in the future, but I'd suggest removing the `p4hir-iso-json` part in this PR.

tools/p4mlir-export/json_emitter.h
```diff
@@ -0,0 +1,41 @@
...
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/Operation.h"
+#include "llvm/Support/JSON.h"
+
+namespace P4::P4MLIR {
+
+class JsonEmitter {
+public:
+
+    // Emit a P4HIR-ISO json, given MLIR module.
```
qobilidop, line 25: As explained above, let's not invent this P4HIR-ISO json format here. Instead, try to emit the BMv2 JSON format directly.

I'd suggest that you only keep the `emitModule` function, and start with returning an empty or constant JSON value. Once you have that working (in the sense that the added test passes, see another comment), we can discuss the next steps.

tools/p4mlir-export/simplebmv2.mlir
qobilidop, line 1: Please add a test in the `test/` directory that invokes the binary on this file as input.

tools/p4mlir-export/simplebmv2.p4
qobilidop: @Joejiong Could you also check in the expected BMv2 `.json` file? I still don't see it.

qobilidop (changes requested):

qobilidop:
> This code only served as the first part of the two-step JSON paths.
p4hir-> p4hir-iso-json (debug+correctness+extensibililty)->bmv2-json (deal with the p4c-bmv2 json generator logic, with ID and others)

@Joejiong I think it makes sense to have multiple steps in converting from P4HIR to BMv2 JSON, but:
1. For this PR, we are not solving that problem. I'd keep the scope of this PR to be just setting up some starter code without assuming how to do the conversion.
2. I don't think it's necessary to introduce an intermediate textual format for solving the conversion problem. We are planning to create a BMv2 dialect for this instead.
3. Without introducing any intermediate steps, in the starter code, we are going to directly convert some easy parts of P4HIR to BMv2 JSON directly. To see which parts are easy, we need to see the expected BMv2 JSON file first. If there is no easy parts, outputing a skeleton BMv2 JSON file is also file.

tools/p4mlir-export/simplebmv2.p4
Joejiong: OK, let me include it

tools/p4mlir-export/bool_test.json
Joejiong: the logic for direct transfer needs more understanding of BMV2 JSON. For now, I could only mimic the BMV2 JSON. Is that what you're saying?

.gitignore
Joejiong: Can I keep it here, just one line, More portable for Mac users. In case, other developers also use MAC, :)

tools/CMakeLists.txt
Joejiong: I feel it is a folder with options, inside this folder different emitters could emit protobuf, JSON, or even p4info, or bmv2 better?

tools/p4mlir-export/simplebmv2.p4
Joejiong: > @Joejiong Could you also check in the expected BMv2 `.json` file? I still don't see it.

The output also has a p4i file, I probably only upload the `.json`

qobilidop: Yes, including `.json` file is enough.

.gitignore
qobilidop: My point is that for people developing on Mac, it makes more sense to include this in their global `gitignore` file. Actually, I'd suggest including all of https://github.com/github/gitignore/blob/main/Global/macOS.gitignore in their global `gitignore`. For this project-wise `gitignore`, I'd prefer keeping it as clean as possible.

tools/p4mlir-export/bool_test.json
qobilidop: I can further explain this once we have the expected BMv2 JSON file checked in.

tools/CMakeLists.txt
Joejiong: got you

.gitignore
Joejiong: > My point is that for people developing on Mac, it makes more sense to include this in their global `gitignore` file. Actually, I'd suggest including all of https://github.com/github/gitignore/blob/main/Global/macOS.gitignore in their global `gitignore`. For this project-wise `gitignore`, I'd prefer keeping it as clean as possible.

let take look at this

tools/p4mlir-export/simplebmv2.json
qobilidop, line 1: Now that we have this expected BMv2 JSON file, we can discuss the next step. The exercise I'd like to see you implement in this PR is to generate the `"header_types"` section automatically.

The implementation shall be very straight-forward. You just need to traverse the input P4HIR program in some way and collect all header type related definitions. Make your best guesses at what is counted as header types. Reference the existing P4C BMv2 backend code for inspirations.

You don't have to generate code that look exactly the same. If they look roughly the same, that's good enough for this PR.

@Joejiong Any questions?

Joejiong: header_types is like a struct declaration？ Got to look at it later, EST time is late. Thx for the elaboration.

qobilidop: No problem. Take your time. https://github.com/p4lang/behavioral-model/blob/main/docs/JSON_format.md has some explanations.

.gitignore
fruffy: Actually, I think it is okay to add this here. Eventually, someone with a MacOS won't have this global setup and then the file is inadvertently added in some bulk commit. It's a single line in the git diff so easy to overlook.

https://github.com/p4lang/p4c/blob/main/.gitignore#L40 has some categorization that can be useful. 

Probably have to add all the dot dirs for the AI tools soon, too :).

qobilidop: Okay, let's keep this then.

Joejiong: Done

tools/CMakeLists.txt
Joejiong: Done

tools/p4mlir-export-bmv2/p4mlir-export-bmv2.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "mlir/Parser/Parser.h"
+#include "p4mlir/Dialect/P4HIR/P4HIR_Dialect.h"
+#pragma GCC diagnostic pop
+
+#include "json_emitter.h"
+
+
+int main(int argc, char *const argv[]) {
+
+    P4::AutoCompileContext autoP4MLIRExportContext(new P4::MLIR::ExportContext);
```
asl, line 32: This should be entirely MLIR / LLVM-based tool, therefore no P4C code should be imported / used.

Joejiong: Thanks, let me change it.

asl: Overall, try not to blindly copy & paste pieces from different places.

Joejiong: This is the draft version, which will later be cleaned out.

Joejiong: > Overall, try not to blindly copy & paste pieces from different places.

Oh, this code is useful; without it, the test will not pass.

asl: Well, maybe you can do one step further? And try to figure out why "the test will not pass" and fix the code afterwards? Provided that your input is pure MLIR.

Joejiong: Okay, got you.

fruffy: >This should be entirely MLIR / LLVM-based tool, therefore no P4C code should be imported / used.

Are you sure? Don't we still need the parser/frontend for the P4 BMv2 program?

asl: Why? The input should be MLIR (in P4HIR dialect) as produced by `p4mlir-translate`. So, no frontend / parser here, only "dialect conversion".

.github/workflows/build-and-test.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   build-and-test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
```
Joejiong, line 17: My build script has been hanging for a long time, so I have split the build file now.

tools/p4mlir-export-bmv2/p4mlir-export-bmv2.cpp
fruffy: Ah, missed the context in the original issue (https://github.com/p4lang/p4mlir-incubator/issues/151#issuecomment-2768081285). Makes sense then.

.github/workflows/build-and-test.yml
qobilidop: Could you rebase or merge with HEAD? It's already updated there.

Joejiong: > Could you rebase or merge with HEAD? It's already updated there.

Yes, later I will rebase and squash the PR I'm not trying to merge the file; it is to test the build system. Somehow, the build system is not working for this PR.

build_tools/unittest.sh
```diff
@@ -0,0 +1,18 @@
...
+
+P4C_REPO_DIR=$P4MLIR_REPO_DIR/third_party/p4c
+P4C_BUILD_DIR=$P4C_REPO_DIR/build
+P4C_EXT_DIR=$P4C_REPO_DIR/extensions
+
+P4C_P4MLIR_EXT_DIR=$P4C_EXT_DIR/p4mlir
+
+cd "$P4C_BUILD_DIR"
+# Run some tests
+ninja check-p4mlir
```
qobilidop, line 15: For the purpose of running unit test, you can simply go to the build directory and run `ninja check-p4mlir`. There is no need for a separate script.

This is also written in the project README already.

Joejiong: Just to test the build system to see why it hangs, not trying to merge the file, just separate the file. Local build and test are fine.

.github/workflows/build-and-test.yml
```diff
@@ -66,3 +66,7 @@ jobs:
     - name: Build and test P4MLIR
       run: |
         build_tools/build_p4c_with_p4mlir_ext.sh
+    
+    - name: Run P4MLIR tests
+      run: |
+        build_tools/unittest.sh
```
qobilidop, line 72: If your goal is to change the CI workflow, I'd suggest do it in a separate PR.

Also, I'd like to know why you might want to do this?

build_tools/unittest.sh
qobilidop: Cool. Thanks for the explanation. That's fine!

Joejiong:
I finally found the error, but it stopped at my newly added test. Why could I run it on my local machine?
```
/mlirTojson/p4mlir-incubator/third_party/llvm-project/build/bin/llvm-lit -v /mlirTojson/p4mlir-incubator/third_party/p4c/build/extensions/p4mlir/test/Export/P4HIR/bool_test.mlir
-- Testing: 1 tests, 1 workers --
PASS: P4MLIR :: Export/P4HIR/bool_test.mlir (1 of 1)

Testing Time: 0.14s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
+ PASSED_TESTS=35
+ echo PASSED
PASSED
```

qobilidop:
In general your local environment and CI environment are not exactly the same. So things could be different. I usually use Docker to create a reproducible local environment that's very close to the CI environment to rule out such possibilities.

Other than that, here are my suggestions on what you could try to debug:

1. Rebase or merge to HEAD to rule out the possibility that something already fixed in HEAD has been causing your issue.
2. Replace the test file by an existing passing test.
   1. If this works in CI, it means your added test has some issue.
   2. If this also doesn't work in CI, then it must be something else.

tools/p4mlir-export-bmv2/p4mlir-export-bmv2.cpp
qobilidop: @Joejiong Anything under the `P4::` namespace comes from P4C. If your input is a `.mlir` program, you won't need that. I'd suggest that you take a look at [`circt-translate`](https://github.com/llvm/circt/blob/main/tools/circt-translate/circt-translate.cpp) and follow the `--export-systemc` option to learn how to build the MLIR binary we need here.

asl:
> If you are referring to thread-local storage with TLS BDWGC should support that. Probably just an option to tweak.

No, it does not. See https://github.com/ivmai/bdwgc/blob/master/README.md:
> WARNING: the collector does not guarantee to scan thread-local storage (e.g. of the kind accessed with pthread_getspecific). The collector does scan thread stacks, though, so generally the best solution is to ensure that any pointers stored in thread-local storage are also stored on the thread's stack for the duration of their lifetime. (This is arguably a longstanding bug, but it hasn't been fixed yet.)

fruffy:
> > If you are referring to thread-local storage with TLS BDWGC should support that. Probably just an option to tweak.
> 
> No, it does not. See https://github.com/ivmai/bdwgc/blob/master/README.md:
> 
> > WARNING: the collector does not guarantee to scan thread-local storage (e.g. of the kind accessed with pthread_getspecific). The collector does scan thread stacks, though, so generally the best solution is to ensure that any pointers stored in thread-local storage are also stored on the thread's stack for the duration of their lifetime. (This is arguably a longstanding bug, but it hasn't been fixed yet.)

It works in practice, but the guarantees aren't good, see https://github.com/ivmai/bdwgc/blob/ef01762da61fc803f430751506f6ee2176fe9221/docs/platforms/README.linux#L59.

asl:
> It works in practice, but the guarantees aren't good, see https://github.com/ivmai/bdwgc/blob/ef01762da61fc803f430751506f6ee2176fe9221/docs/platforms/README.linux#L59.

Well, I definitely saw crashes and double frees with MLIR with GC enabled :) So, I disabled it by purpose.

fruffy:
> Well, I definitely saw crashes and double frees with MLIR with GC enabled :) So, I disabled it by purpose.

I guess the question is, does `GC_disable()` work properly? Or did you have to disable GC entirely? The MLIR code doesn't need it anyway and we only really need GC for the front-end for large programs.

asl:
> I guess the question is, does `GC_disable()` work properly?

It stops all mark&sweep threads, so it means no automatic memory frees anymore. Manual allocation / deallocation work (albeit slow).

> need GC for the front-end for large programs.
Sadly, "large" is just couple of thousands of LOCs, if entire frontend is run :)

fruffy:
> It stops all mark&sweep threads, so it means no automatic memory frees anymore. Manual allocation / deallocation work (albeit slow).

So, in the case of the translate utility, if we call `GC_disable()` after the front end passes, the only leakage we should have is the P4 Program plus whatever leaks from the call to `toMLIR`, right?

asl:
> So, in the case of the translate utility, if we call `GC_disable()` after the front end passes, the only leakage we should have is the P4 Program plus whatever leaks from the call to `toMLIR`, right?

plus whatever is in the static variables of the frontend, some some data structures (we still rely on type map, sadly in few places), plus whatever is not cleared after the frontend, yes.

qobilidop:
@asl @fruffy Is the discussion above settled? Or would it be useful to summarize the discussion so far in an issue for future reference and further discussion?

PR #155, opened 2025-04-08
Make ReturnOp take Optional operands

xl4624:
Small thing I noticed while starting on the `continue` and `break` operations.

https://mlir.llvm.org/docs/DefiningDialects/Operations/#optional-operands

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
@@ -760,18 +760,18 @@ def ReturnOp : P4HIR_Op<"return", [ParentOneOf<["ScopeOp", "IfOp",
 
   // The return operation takes an optional input operand to return. This
   // value must match the return type of the enclosing function.
-  let arguments = (ins Variadic<AnyP4Type>:$input);
+  let arguments = (ins Optional<AnyP4Type>:$input);
 
   // The return operation only emits the input in the format if it is present.
-  let assemblyFormat = "($input^ `:` type($input))? attr-dict ";
+  let assemblyFormat = [{ ($input^ `:` type($input))? attr-dict }];
 
   // Allow building a ReturnOp with no return operand.
   let builders = [
-    OpBuilder<(ins), [{ build($_builder, $_state, std::nullopt); }]>
+    OpBuilder<(ins), [{ build($_builder, $_state, nullptr); }]>
   ];
 
   let extraClassDeclaration = [{
-    bool hasOperand() { return getNumOperands() != 0; }
+    bool hasOperand() { return getInput() != nullptr; }
   }];
 
   let hasVerifier = 1;
@@ -1600,7 +1600,6 @@ let description = [{
 
   let hasCustomAssemblyFormat = 1;
 }
-
 
 def ConditionOp : P4HIR_Op<"condition",
         [HasParent<"ForOp">,
--- a/lib/Dialect/P4HIR/P4HIR_Ops.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Ops.cpp
@@ -644,7 +644,7 @@ void P4HIR::IfOp::build(OpBuilder &builder, OperationState &result, Value cond, 
     elseBuilder(builder, result.location);
 }
 
-mlir::LogicalResult P4HIR::ReturnOp::verify() {
+LogicalResult P4HIR::ReturnOp::verify() {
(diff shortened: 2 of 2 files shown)


include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
```diff
@@ -726,17 +726,12 @@ def IfOp : P4HIR_Op<"if",
   ];
 }
 
-def ReturnOp : P4HIR_Op<"return", [ParentOneOf<["ScopeOp", "IfOp",
```
xl4624, line 729: Is there any reason why we explicitly check for `ParentOneOf` here? Curious why is the verifier checking that we are in either a `function`, `action`, or `control` not enough.

asl: These are not direct parents (well, they are now as we using `RemoveReturns` pass that eliminates all nested returns), so we cannot check that we are there.

xl4624: Thanks for the explanation and link to the thread @asl

> These are not direct parents

Just to make sure I understand this correctly, by "These" are u referring to `p4hir.func` and `p4hir.control`, as in they won't always be the direct parent of `p4hir.return`? For example:
```
p4hir.func @test() {
	p4hir.if %cond {
		p4hir.return
	}
	...
}
```
Here, `ParentOneOf` would be checking against the operation containing the region's block (like `p4hir.if`), not necessarily the top-level `p4hir.func`, right?

My thinking was, isn't the fundamental semantic restriction that a `p4hir.return` must ultimately belong _somewhere_ inside a function, action, or control, regardless of its immediate parent? Can't we relax the immediate parent restriction and just rely on:

```cpp
LogicalResult P4HIR::ReturnOp::verify() {
	auto fnOp = getOperation()->getParentOfType<FunctionOpInterface>();
	if (!fnOp || !mlir::isa<P4HIR::FuncOp, P4HIR::ControlOp>(fnOp)) {
	    return emitOpError() << "returns are only possible from function-like objects: functions, "
	                            "actions and control apply blocks";
}
```

asl: > Here, ParentOneOf would be checking against the operation containing the region's block (like p4hir.if), not necessarily the top-level p4hir.func, right?

Yes, exactly.

> Can't we remove the immediate parent restriction and only rely on:

There are multitude of the issues outlined in the link above, e.g.:
 - Do we want return to be a terminator? If yes, then what to do with dead code after return?
 - Same, but for `ReturnLike` trait.

Overall, MLIR favors to SEME regions, as a result we cannot have, say, `return` in the middle of the block (same problem with `break` and `continue` just in case) as terminators and return-like instructions only transfer execution flow to the parent region.

Previously we used to have `return` operation that was not terminator (and therefore if very unclear control-flow semantics) and `implicit_return` terminator instruction. This was pretty awkward and I solved the problem via running P4-level `RemoveReturns` pass. It has its own issues and quite inefficient, but for now solves the problem.

xl4624: Ah, I think I understand now, so for now we rely on `RemoveReturn` to simplify control flow via a flag, but the idea is that once that RFC lands and region-based control flow is available in MLIR, we can remove that frontend pass and use their control flow semnatics to keep the HIR closer to the original syntax.

include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
```diff
@@ -726,17 +726,12 @@ def IfOp : P4HIR_Op<"if",
...
-                                                "FuncOp", "ControlApplyOp",
-                                                "CaseOp",
-                                                "ForOp", "ForInOp",
-                                             ]>,
-                                             ReturnLike, Terminator
+def ReturnOp : P4HIR_Op<"return",
+  [ReturnLike, Terminator]> {
```
xl4624, line 730:
```suggestion
  [ParentOneOf<["FuncOp", "ControlApplyOp"]>,
   ReturnLike, Terminator]> {
```

Since now returns are terminators, can we narrow back down to this?

asl: Let's keep it as-is for now :)

xl4624:
Reverted the `ParentOneOf` change but I don't really have a strong opinion on the rest of the PR as they feel more like nits.

PR #157, opened 2025-04-08
Use Ubuntu 24.04 in CI

asl:


fruffy:
I prefer pinning the Ubuntu version to avoid having to deal with breaking changes.

asl:
> I prefer pinning the Ubuntu version to avoid having to deal with breaking changes.

Well, I usually prefer the other way round: if there are some incompatibilities it is better to know them sooner than later.

fruffy:
> > I prefer pinning the Ubuntu version to avoid having to deal with breaking changes.
> 
> Well, I usually prefer the other way round: if there are some incompatibilities it is better to know them sooner than later.

The problem is when these occur at inconvenient times and end up blocking a PR. I had to deal with this a lot on P4C, so I prefer pinning these days. 

Although we probably do not have to worry about this here for now, our test matrix is not yet complex.

fruffy (approved):
This PR might likely break because of Ubuntu 24.04 and pip's introduction of `--break-system-packages`.

asl:
Yeah, we will see.

qobilidop:
Ubuntu was pinned to 20.04 version mainly to make sure P4C builds. It seems P4C now also builds with newer Ubuntu versions. Agree we could try and see if there are any issues. No other concerns from my side.

qobilidop:
Just found some related discussion on the P4C side: https://github.com/p4lang/p4c/issues/5155. @fruffy Does P4C supports Ubuntu 24.04 now, or are there still known blocking issues?

fruffy:
> Just found some related discussion on the P4C side: [p4lang/p4c#5155](https://github.com/p4lang/p4c/issues/5155). @fruffy Does P4C supports Ubuntu 24.04 now, or are there still known blocking issues?

There are still some issues but these are related to the test frameworks. This repository should have no problems compiling and I do not think it uses any of the Python test infra. Once we run BMv2 we have to worry about that.

qobilidop (approved):

fruffy:
I am actually surprised this passes. I am running Ubuntu 24.04 and this code here https://github.com/p4lang/p4mlir-incubator/blob/main/build_tools/ubuntu_install_mlir_requirements.sh#L19 causes some problems. 

What CMake version do we actually need?

qobilidop:
@fruffy LLVM/MLIR requires `CMake >=3.20.0` (https://llvm.org/docs/GettingStarted.html#software).

fruffy:
> @fruffy LLVM/MLIR requires `CMake >=3.20.0` (https://llvm.org/docs/GettingStarted.html#software).

iirc Ubuntu 22.04 comes with cmake 3.20+ so we can remove the pip line, will do a PR later.

PR #158, opened 2025-04-08
improve build & test speed

Joejiong:
The old building speed is a little slow, I don't know if I could do this increase in speed for the first time, when the cache did not hit at first build.

### DIFF
--- a/build_tools/build_mlir.sh
+++ b/build_tools/build_mlir.sh
@@ -40,6 +40,8 @@ CMAKE_FLAGS+=" -DLLVM_ENABLE_EH=ON"
 
 cmake -G Ninja "$LLVM_REPO_DIR"/llvm $CMAKE_FLAGS
 
-ninja
+JOBS=$(nproc 2>/dev/null || sysctl -n hw.ncpu)
+ninja -j$JOBS
+
 ninja check-mlir
 ninja install
--- a/build_tools/build_p4c_with_p4mlir_ext.sh
+++ b/build_tools/build_p4c_with_p4mlir_ext.sh
@@ -57,7 +57,9 @@ CMAKE_FLAGS+=" -DLLVM_EXTERNAL_LIT=$LLVM_BUILD_DIR/bin/llvm-lit"
 mkdir -p "$P4C_BUILD_DIR"
 cd "$P4C_BUILD_DIR"
 cmake -G Ninja "$P4C_REPO_DIR" $CMAKE_FLAGS
-ninja
+
+JOBS=$(nproc 2>/dev/null || sysctl -n hw.ncpu)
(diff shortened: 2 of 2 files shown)


build_tools/build_mlir.sh
```diff
@@ -40,6 +40,8 @@ CMAKE_FLAGS+=" -DLLVM_ENABLE_EH=ON"
 
 cmake -G Ninja "$LLVM_REPO_DIR"/llvm $CMAKE_FLAGS
 
-ninja
+JOBS=$(nproc 2>/dev/null || sysctl -n hw.ncpu)
+ninja -j$JOBS
```
asl, line 44: Ninja defaults to all CPU cores available, this is not `make`.

asl (changes requested):

build_tools/build_mlir.sh
Joejiong: > Ninja defaults to all CPU cores available, this is not `make`.

I heard it is not the maximum, I don't know if I manually set it to available, which could make a difference. Let me double-check it.

Joejiong: I heard that the difference is probably available or preemptive. Probably, when other jobs are running, this can grab resources, just a guess. But, you are right, it should be the maximum cores available.

fruffy: Yeah, on systems with constrained resources ninja's approach can lead to annoying lockups. It can be advisable to use one core less than the max. But your change is functionally equivalent right now.

asl: @fruffy It is more interesting actually. On my MBP laptop I'm having total 14 cores, but among those 10 performance and 4 efficiency. So I can go `-j 14`, but in the reality it does not make sense to go beyond `-j 10` as I really see difference in build speed in performance vs efficiency cores. It's better to keep them for "normal" tasks.

Joejiong: Actually, I checked that the Ubuntu runner only has 2 cores. :0

```
2 cores for standard GitHub-hosted runners
Processor: x86_64
Memory: Approximately 7 GB of RAM
```
We can add those to the action workflow.
```
- name: Detailed System Information
  run: lscpu
```

asl: This is public information: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

No need to guess

fruffy: > Actually, I checked that the GitHub action Ubuntu runner only has 2 cores. :0

They should have four cores, 2 if you are using a private repository.

PR #160, opened 2025-04-11
Update Ubuntu install scripts.

fruffy:
Don't use `pip`, it can break on newer versions of Ubuntu.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -14,10 +14,13 @@ concurrency:
 
 jobs:
   build-and-test:
-    runs-on: ubuntu-24.04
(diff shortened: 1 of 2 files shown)


qobilidop (approved):

qobilidop:
Weird it's still expecting the previous `build-and-test` CI to pass. Maybe it's something in our repo settings?

qobilidop:
image

Yeah we currently have this. I've temporarily removed the requirement on `build-and-test`.

qobilidop:
> I've temporarily removed the requirement on `build-and-test`.

And added it back. The required check is still called `build-and-test`. So I guess we could have force-merged this PR as a simpler solution.

PR #162, opened 2025-04-13
Refine ccache key setting to involve version of runner OS and changes…

tagolog:
… in the MLIR build script

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -14,11 +14,12 @@ concurrency:
 
 jobs:
   build-and-test:
-    name: test-${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+    name: test-${{ matrix.distribution }}
+    runs-on: ${{ matrix.distribution }}
     strategy:
+      fail-fast: false  # Let all matrix jobs complete, even if one fails
       matrix:
-        os: [ubuntu-22.04, ubuntu-24.04]
+        distribution: [ubuntu-22.04, ubuntu-24.04]
 
     steps:
     - name: Checkout repo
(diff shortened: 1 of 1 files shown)


fruffy (commented):
Would also add `fail-fast: false` to the strategy. And rename from `matrix.os` to `matrix.distribution`.

`llvm-commit-${{ steps.get-llvm-hash.outputs.hash }}-mlir-build-script-${{ hashFiles('${{ github.workspace }}/build_tools/build_mlir.sh') }}` Do we actually need all this? Might become a maintenance headache. If the cache worked before it should work with the two different dist types.

tagolog:
> Would also add `fail-fast: false` to the strategy. And rename from `matrix.os` to `matrix.distribution`.

@fruffy  Thanks for the suggestions!
- fail-fast: false - added.
- "matrix.os" renamed to "matrix.distribution"

> Do we actually need all this?
I think we need these parts of the key.
- llvm-commit-{hash} - Required because LLVM's codebase can change between commits. Without this, a cache hit could return incompatible build results.
mlir-build-script-{hash} - The build_mlir.sh script defines critical build flags (e.g., -DLLVM_ENABLE_RTTI=ON, compiler choices, linker options). If the script changes (e.g., enabling/disabling features), cached objects from a previous version might be invalid or unsafe to reuse.

The comment was appended to the "build-and-test.yml" to clarify key parts.

@asl what do you think?

fruffy (approved):
> > Would also add `fail-fast: false` to the strategy. And rename from `matrix.os` to `matrix.distribution`.
> 
> @fruffy Thanks for the suggestions!
> 
>     * fail-fast: false - added.
> 
>     * "matrix.os" renamed to "matrix.distribution"
> 
> 
> > Do we actually need all this?
> 
> I think we need these parts of the key.
> 
>     * llvm-commit-{hash} - Required because LLVM's codebase can change between commits. Without this, a cache hit could return incompatible build results.
> 
>     * mlir-build-script-{hash} - The build_mlir.sh script defines critical build flags (e.g., -DLLVM_ENABLE_RTTI=ON, compiler choices, linker options). If the script changes (e.g., enabling/disabling features), cached objects from a previous version might be invalid or unsafe to reuse.
> 
> 
> The comment was appended to the "build-and-test.yml" to clarify key parts.
> 
> @asl what do you think?

I just saw that we are also caching the `install` directory. In that case it makes sense. I thought we are only caching `ccache`-related products.

tagolog:
@fruffy , @asl can you please merge this PR? I have no merge permissions.

Issue #28, opened 2025-01-24
Implement folders for common canonicalization


asl:
More information: https://mlir.llvm.org/docs/Canonicalization/

asl:
* Fold trivial casts
* Canonicalize read + struct_extract into struct_extract_ref + read as a simple SROA
* Struct + struct_extract => just single field

PR #163, opened 2025-04-14
Add basic unary folds

xl4624:
Implements https://github.com/p4lang/p4mlir-incubator/issues/28#issuecomment-2613585887

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
@@ -294,6 +294,7 @@ def UnaryOp : P4HIR_Op<"unary",
   }];
 
   let hasVerifier = 1;
+  let hasFolder = 1;
 }
 
 def BinOpKind_Mul    : I32EnumAttrCase<"Mul",   1, "mul">;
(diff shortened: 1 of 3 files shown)


xl4624:
Also is it normal for p4c local builds to take 5 minutes to complete? I was looking into the tc/ebpf unparsed annotations issue to get familiar but I probably spent 30 minutes alone just rebuilding over and over :/

fruffy:
> Also is it normal for p4c local builds to take 5 minutes to complete? I was looking into the tc/ebpf unparsed annotations issue to get familiar but I probably spent 30 minutes alone just rebuilding over and over :/

Full P4C build takes a while. It should compile quickly afterwards unless you are modifying core files. Do you have `ccache`?

xl4624:
Ended up switching to the Dockerfile instead. I'm on an Arch laptop and had some tests failing like this:
```
In file included from /usr/include/stdio.h:437:
/usr/include/bits/floatn.h:83:52: error: unsupported machine mode '__TC__'
typedef _Complex float __cfloat128 __attribute__ ((__mode__ (__TC__)));
                                                   ^
/usr/include/bits/floatn.h:97:9: error: __float128 is not supported on this target
typedef __float128 _Float128;
```
Seems to work fine now. Thanks!

asl:
> I was looking into the tc/ebpf unparsed annotations issue to get familiar but I probably spent 30 minutes alone just rebuilding over and over :/

These are nasty. And probably not worth looking as the code around is a bit convoluted. That said, I do rely on `ccache` for faster compile times.

asl (approved):
Great, thanks!

PR #165, opened 2025-04-15
Add basic shift folds

xl4624:
I renamed `verifyArithmeticShiftOperation()` -> `verifyShiftOperation()` as well as the corresponding `OpError` message because, from my understanding from the spec, logical and arithmetic shifts depend on whether the left operand's type is signed or unsigned, and since we check this for unsigned integers as well the error shouldn't specify "arithmetic shift" (?)

Also question: can we fold `InfIntType` shifts? I initially was going to handle constant `InfIntType` lhs values by applying the formula from the spec:

> The expression `a << b` is equal to `a x 2^b` while `a >> b` is equal to `a / 2^b`

But I saw that for example [CastOp::fold()](https://github.com/p4lang/p4mlir-incubator/blob/80e796ce6bac2045941eb1ef6285ff7a13e3052f/lib/Dialect/P4HIR/P4HIR_Ops.cpp#L170) only folds if the destination type is a fixed-width integer. Is that just a `TODO` or are there limitations on folding `InfIntType`s? I'm guessing the latter but wanted to confirm just in case, Thanks!

### DIFF
--- a/lib/Dialect/P4HIR/P4HIR_Ops.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Ops.cpp
@@ -251,7 +251,7 @@ OpFoldResult P4HIR::UnaryOp::fold(FoldAdaptor adaptor) {
     if (getKind() == P4HIR::UnaryOpKind::UPlus) return getInput();
 
     // Double negation.
-    if (auto inputOp = mlir::dyn_cast_or_null<UnaryOp>(getInput().getDefiningOp())) {
+    if (auto inputOp = mlir::dyn_cast_if_present<UnaryOp>(getInput().getDefiningOp())) {
         if (getKind() == inputOp.getKind()) {
             if (getKind() == P4HIR::UnaryOpKind::LNot ||  // not(not(%a)) -> %a
                 getKind() == P4HIR::UnaryOpKind::Cmpl ||  // compl(compl(%a)) -> %a
@@ -291,7 +291,7 @@ void P4HIR::BinOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
 }
 
 //===----------------------------------------------------------------------===//
-// ShrOp, ShlOp
+// ShrOp & ShlOp
 //===----------------------------------------------------------------------===//
 
 void P4HIR::ShlOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
@@ -302,13 +302,84 @@ void P4HIR::ShrOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
     setNameFn(getResult(), "shr");
 }
 
+LogicalResult verifyShiftOperation(Operation *op, Type rhsType) {
+    if (auto rhsBitsType = dyn_cast<P4HIR::BitsType>(rhsType)) {
+        if (rhsBitsType.isSigned()) {
+            return op->emitOpError("the right-hand side operand of a shift must be unsigned");
+        }
+    }
+    return success();
+}
+
+LogicalResult P4HIR::ShlOp::verify() {
+    auto rhsType = getRhs().getType();
+    return verifyShiftOperation(getOperation(), rhsType);
+}
+
+LogicalResult P4HIR::ShrOp::verify() {
+    auto rhsType = getRhs().getType();
+    return verifyShiftOperation(getOperation(), rhsType);
+}
+
+OpFoldResult P4HIR::ShlOp::fold(FoldAdaptor adaptor) {
+    // Can't fold anything if shift amount is not constant
+    auto rhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getRhs());
+    if (!rhsAttr) return {};
+    auto shift = rhsAttr.getUInt();
+
+    // Identity.
+    // shl(%x, 0) -> %x
+    if (shift == 0) return getLhs();
+
+    if (auto bitsType = mlir::dyn_cast<P4HIR::BitsType>(getType())) {
+        auto width = bitsType.getWidth();
+        // shl(%x : bit/int<W>, c) -> 0 if c >= W
+        if (shift >= width) return P4HIR::IntAttr::get(bitsType, APInt::getZero(width));
+
+        if (auto lhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getLhs())) {
+            // Arithmetic and logical left shifts produce the same value
+            return P4HIR::IntAttr::get(bitsType, lhsAttr.getValue().shl(shift));
+        }
+    }
+
+    return {};
+}
+
+OpFoldResult P4HIR::ShrOp::fold(FoldAdaptor adaptor) {
+    // Can't fold anything if shift amount is not constant
+    auto rhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getRhs());
+    if (!rhsAttr) return {};
+    auto shift = rhsAttr.getUInt();
+
+    // Identity.
+    // shl(%x, 0) -> %x
+    if (shift == 0) return getLhs();
+
+    if (auto bitsType = mlir::dyn_cast<P4HIR::BitsType>(getType())) {
+        auto width = bitsType.getWidth();
+        bool isSigned = bitsType.isSigned();
+        // shr(%x : bit<W>, c) -> 0 if c >= W
+        if (!isSigned && shift >= width) {
+            return P4HIR::IntAttr::get(bitsType, APInt::getZero(width));
+        }
+
+        if (auto lhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getLhs())) {
+            auto result =
+                isSigned ? lhsAttr.getValue().ashr(shift) : lhsAttr.getValue().lshr(shift);
+            return P4HIR::IntAttr::get(getContext(), bitsType, result);
+        }
+    }
+
+    return {};
+}
+
 //===----------------------------------------------------------------------===//
 // ConcatOp
 //===----------------------------------------------------------------------===//
 
 LogicalResult P4HIR::ConcatOp::verify() {
-    auto lhsType = cast<BitsType>(getOperand(0).getType());
-    auto rhsType = cast<BitsType>(getOperand(1).getType());
+    auto lhsType = cast<BitsType>(getLhs().getType());
+    auto rhsType = cast<BitsType>(getRhs().getType());
     auto resultType = cast<BitsType>(getResult().getType());
 
     auto expectedWidth = lhsType.getWidth() + rhsType.getWidth();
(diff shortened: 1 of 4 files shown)


test/Transforms/Folds/unary.mlir
```diff
@@ -1,72 +1,82 @@
...
+#int-5_b8i = #p4hir.int<-5> : !b8i
+#int-6_b8i = #p4hir.int<-6> : !b8i
+#true = #p4hir.bool<true> : !b
+#false = #p4hir.bool<false> : !b
 
+// CHECK-LABEL: module
 module {
-  p4hir.func @blackhole(!b8)
+
+  // CHECK: %[[true:.*]] = p4hir.const #true
```
xl4624, line 15: Cleaning up some tests

asl:
> But I saw that for example [CastOp::fold()](https://github.com/p4lang/p4mlir-incubator/blob/80e796ce6bac2045941eb1ef6285ff7a13e3052f/lib/Dialect/P4HIR/P4HIR_Ops.cpp#L170) only folds if the destination type is a fixed-width integer. Is that just a `TODO` or are there limitations on folding `InfIntType`s? I'm guessing the latter but wanted to confirm just in case, Thanks!

yes, likely a missed case. `InfInt` are infinite precision integers, so we can fold them as well, no problem

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -327,24 +315,128 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+void P4HIR::ShlOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
+    setNameFn(getResult(), "shl");
+}
+
+void P4HIR::ShrOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
+    setNameFn(getResult(), "shr");
+}
+
+LogicalResult verifyShiftOperation(Operation *op, Type rhsType) {
+    // FIXME: Relax this condition for compile-time known non-negative values.
```
xl4624, line 327: From the spec:

> [T]he right operand must be either an unsigned number of type bit\<S\> or a compile-time known value that is a non-negative integer.

I could be missing something from the frontend pass that already handles this, not sure

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -173,7 +173,7 @@ OpFoldResult P4HIR::CastOp::fold(FoldAdaptor) {
     if (getOperand().getType() == getType()) return getOperand();
 
     // Casts of integer constants
-    if (auto inputConst = mlir::dyn_cast_or_null<ConstOp>(getOperand().getDefiningOp())) {
+    if (auto inputConst = mlir::dyn_cast_if_present<ConstOp>(getOperand().getDefiningOp())) {
```
xl4624, line 176: Saw this comment for `dyn_cast_or_null`:
> // Forwards to dyn_cast_if_present to avoid breaking current users. This is
// deprecated and will be removed in a future patch, use
// dyn_cast_if_present instead.

xl4624:
@asl - could you review when you have the chance? Thanks

lib/Dialect/P4HIR/P4HIR_Ops.cpp
asl: yeah, old habit :)

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -327,24 +315,135 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+    auto rhsType = getRhs().getType();
+    return verifyShiftOperation(getOperation(), rhsType);
+}
+
+template <typename ShiftOp>
+OpFoldResult foldZeroConstants(ShiftOp op, typename ShiftOp::FoldAdaptor adaptor) {
+    // Identity.
+    // shl/shr(%x, 0) -> %x
+    if (auto rhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getRhs())) {
+        if (rhsAttr.getUInt() == 0) return op.getLhs();
```
asl, line 351: I would use `APInt::isZero()` here. Same below.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -327,24 +315,135 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+    }
+
+    // Zero.
+    // shl/shr(0, c) -> 0
+    if (auto lhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getLhs())) {
+        if (lhsAttr.getValue() == 0) {
+            if (auto bitsType = mlir::dyn_cast<P4HIR::BitsType>(op.getType())) {
+                return P4HIR::IntAttr::get(op.getType(), APInt::getZero(bitsType.getWidth()));
+            } else if (auto infIntType = mlir::dyn_cast<P4HIR::InfIntType>(op.getType())) {
+                return P4HIR::IntAttr::get(op.getType(), APInt::getZero(1));
```
asl, line 361: Can we preserve the original bitwidth? I think you can get if from `getValue()` and then unify both branches.

xl4624: Just added in latest commit

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -327,24 +315,132 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+}
+
+OpFoldResult P4HIR::ShlOp::fold(FoldAdaptor adaptor) {
+    if (auto fold = foldZeroConstants(*this, adaptor)) {
+        return fold;
+    }
+
+    auto rhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getRhs());
+    if (!rhsAttr) return {};
+    auto shift = rhsAttr.getUInt();
```
asl, line 373: this might assert for "large" infints (larger than 64 bits)?

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -327,24 +315,132 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+}
+
+OpFoldResult P4HIR::ShrOp::fold(FoldAdaptor adaptor) {
+    if (auto fold = foldZeroConstants(*this, adaptor)) {
+        return fold;
+    }
+
+    auto rhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getRhs());
+    if (!rhsAttr) return {};
+    auto shift = rhsAttr.getUInt();
```
asl, line 408: see above

lib/Dialect/P4HIR/P4HIR_Ops.cpp
xl4624: Oops I didn't even think about really large shifts. so how do we handle large shifts during infint folding? is something like:

```cpp
auto shift = rhsAttr.getValue();
// Shift overflow.
// shl(%x : bit/int<W>, c) -> 0 if c >= W
if (auto bitsType = mlir::dyn_cast<P4HIR::BitsType>(getType())) {
    unsigned width = bitsType.getWidth();
    if (shift.uge(width)) return P4HIR::IntAttr::get(bitsType, APInt::getZero(width));
}
...
// Constant folding
if (auto lhsAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(adaptor.getLhs())) {
	//...
	} else if (auto infIntType = mlir::dyn_cast<P4HIR::InfIntType>(getType())) {
		if (!shift.tryZExtValue()) return {};
```

ok?

asl: Just do them using APInt. This should be enough. If `lhs` is a constant, then you can just use APInt methods for constfolding (regardless whether its `BitsType` or `InfInt`).

xl4624: I see, I looked at how P4C does constant folding of shifts, and I realized that they also do a check for when shift would overflow: `shift > maximumWidthSupported()`, which defaults to 2048. Should we also add a check like that, and if so, how do we log the error?

```cpp
std::optional<uint64_t> shiftOpt = shift.tryZExtValue();
if (!shiftOpt || shiftOpt.value() > 2048) {
    // log here?
	return {};
}
uint64_t shiftAmt = shiftOpt.value();
```

Only reason I wanted to unwrap shift to `shiftAmt` is because because `APInt::shl(const APInt&)` clamps shift amount to `bitWidth`. From their source code:
```cpp
APInt &APInt::operator<<=(const APInt &shiftAmt) {
  // It's undefined behavior in C to shift by BitWidth or greater.
  *this <<= (unsigned)shiftAmt.getLimitedValue(BitWidth);
  return *this;
}

```

We already check for this for fixed-sized integers, and it's easy to do while keeping `APInt shift` with `shift.uge(bitsType.getWidth)`, but for infints we would have to extend lhs's value be able to hold the resulting value `lhsVal = lhsVal.sext(lhsVal.getActiveBits() + shiftAmt + 1)` (and `APInt::sext()` takes unsigned int).

xl4624:
@asl addressed PR feedback. Still ended up unwrapping shift to a uint64_t to extend infint lhs bit widths properly (`sextOrTrunc()` takes an unsigned int argument), but added a check first

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -327,24 +315,126 @@ LogicalResult P4HIR::ConcatOp::verify() {
...
+    // shl(%x : bit/int<W>, c) -> 0 if c >= W
+    if (auto bitsType = mlir::dyn_cast<P4HIR::BitsType>(getType())) {
+        unsigned width = bitsType.getWidth();
+        if (shift.uge(width)) return P4HIR::IntAttr::get(bitsType, APInt::getZero(width));
+    }
+
+    // Check if shift amount fits in a uin64_t
+    std::optional<uint64_t> shiftOpt = shift.tryZExtValue();
+    if (!shiftOpt.has_value()) {
+        // TODO: Should we log an error here? P4C throws an error on shift > 2048
```
asl, line 397: It would leave as-is.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -1513,7 +1603,8 @@ void P4HIR::ParserOp::print(mlir::OpAsmPrinter &printer) {
     }
 
     printer << ' ';
-    printer.printRegion(getRegion(), /*printEntryBlockArgs=*/false, /*printBlockTerminators=*/true);
+    printer.printRegion(getRegion(), /*printEntryBlockArgs=*/false,
```
asl, line 1606: Are these changes from clang-format? I think it would fold back to the original line.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
xl4624: Removed the shift TODO + accidental clang-format changes

asl (approved):
Thanks!

Issue #98, opened 2025-03-06
Implement parser control flow simplification


asl:
Similar to `SimplifyParsers` pass

asl:
Implemented in #166

PR #166, opened 2025-04-28
Add SimplifyParsers pass

xl4624:
Implements #98


lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+#define GEN_PASS_DEF_SIMPLIFYPARSERS
+#include "p4mlir/Transforms/Passes.cpp.inc"
+
+namespace {
+struct SimplifyParsers : public impl::SimplifyParsersBase<SimplifyParsers> {
+    void runOnOperation() override;
+
+ private:
+    /// Finds all states reachable from the 'start' state.
+    llvm::DenseSet<P4HIR::ParserStateOp> findReachableStates(P4HIR::ParserOp &parser,
```
asl, line 21: P4C mainline has a way to build "parser call graph". I'm wondering if it would be possible to build something similar here? I have not looked if MLIR callgraph functionality is generic enough for this.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    return visited;
+}
+
+void SimplifyParsers::buildIndegreeAndSuccessor(
+    P4HIR::ParserOp &parser, mlir::SymbolTable &symbolTable,
+    llvm::DenseMap<P4HIR::ParserStateOp, unsigned> &indegree,
+    llvm::DenseMap<P4HIR::ParserStateOp, P4HIR::ParserStateOp> &succ) {
+    // Helper function to identify special states that should never be collapsed
+    auto isSpecial = [](P4HIR::ParserStateOp state) {
```
asl, line 80: name should be meaningless here (e.g. `verify` call is expanded into reject). So, we need to look into states that:
 - Has transition from the body. This will be start
 - Has accept / reject op inside.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+        auto name = state.getSymName();
+        return name == "start" || name == "accept" || name == "reject";
+    };
+
+    for (auto state : parser.getBody().getOps<P4HIR::ParserStateOp>()) {
+        if (isSpecial(state)) continue;
```
asl, line 86: Looks like this is unnecessary. Also, the "start" state is not special anymore. It is the "control flow" that matters.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+        Operation *term = state.getBody().back().getTerminator();
+        mlir::TypeSwitch<Operation *, void>(term)
+            .Case<P4HIR::ParserAcceptOp, P4HIR::ParserRejectOp>([&](auto) {})
+            .Case<P4HIR::ParserTransitionOp>([&](auto tr) {
+                auto next =
```
asl, line 92: This looks common enough. Maybe factor out into some method of ParserStateOp? E.g. iterate over next states / get next state given the name / symbol.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+                    symbolTable.lookup<P4HIR::ParserStateOp>(tr.getStateAttr().getLeafReference());
+                if (isSpecial(next)) return;
```
asl, line 94: See above. Even more, I would not consider accept / reject "states" as something special given we're having dedicated terminator ops. So, we can use them as normal terminators. And if the target would want this, they could always split them back into separate states.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+                ++indegree[next];
+                succ[state] = next;
+            })
+            .Case<P4HIR::ParserTransitionSelectOp>([&](P4HIR::ParserTransitionSelectOp sel) {
+                llvm::SmallVector<P4HIR::ParserStateOp> targets;
+                for (auto selectCase : sel.getBody().getOps<P4HIR::ParserSelectCaseOp>()) {
```
asl, line 101: See above about next states. Technically we can use LLVM GraphTraits to provide graph-like iteration. Note sure how MLIR'ish it is :)

asl (commented):
See comments. I would probably introduce some graph-like API for iterating over next / previous states and use this pass as an opportunity to extend IR as it could be convenient.

include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
```diff
@@ -38,6 +38,9 @@ def ParserOp : P4HIR_Op<"parser",
     }
 
     void createEntryBlock();
+
+    llvm::SmallVector<mlir::Operation *> getSuccessors(P4HIR::ParserStateOp &state,
```
xl4624, line 42: I ended up having to use `llvm::SmallVector<Operation *>` here because of this incomplete type error:

```bash
In file included from /home/xiaomin/dev/p4mlir/third_party/p4c/extensions/p4mlir/tools/p4mlir-opt/p4mlir-opt.cpp:1:
In file included from /home/xiaomin/dev/p4mlir/install/include/mlir/IR/MLIRContext.h:13:
In file included from /home/xiaomin/dev/p4mlir/install/include/mlir/Support/TypeID.h:21:
In file included from /home/xiaomin/dev/p4mlir/install/include/llvm/Support/Allocator.h:20:
/home/xiaomin/dev/p4mlir/install/include/llvm/ADT/SmallVector.h:1172:71: error: invalid application of 'sizeof' to an incomplete type 'P4::P4MLIR::P4HIR::ParserStateOp'
 1172 |   static constexpr size_t NumElementsThatFit = PreferredInlineBytes / sizeof(T);
      |                                                                       ^~~~~~~~~
/home/xiaomin/dev/p4mlir/install/include/llvm/ADT/SmallVector.h:1174:7: note: in instantiation of static data member 'llvm::CalculateSmallVectorDefaultInlinedElements<P4::P4MLIR::P4HIR::ParserStateOp>::NumElementsThatFit' requested here
 1174 |       NumElementsThatFit == 0 ? 1 : NumElementsThatFit;
      |       ^
/home/xiaomin/dev/p4mlir/install/include/llvm/ADT/SmallVector.h:1194:71: note: in instantiation of static data member 'llvm::CalculateSmallVectorDefaultInlinedElements<P4::P4MLIR::P4HIR::ParserStateOp>::value' requested here
 1194 |           unsigned N = CalculateSmallVectorDefaultInlinedElements<T>::value>
      |                                                                       ^
/home/xiaomin/dev/p4mlir/third_party/p4c/build/extensions/p4mlir/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.h.inc:6996:30: note: in instantiation of default argument for 'SmallVector<P4HIR::ParserStateOp>' required here
 6996 |   llvm::iterator_range<llvm::SmallVector<P4HIR::ParserStateOp>> getSuccessors(P4HIR::ParserStateOp &state,
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/xiaomin/dev/p4mlir/third_party/p4c/build/extensions/p4mlir/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.h.inc:230:7: note: forward declaration of 'P4::P4MLIR::P4HIR::ParserStateOp'
  230 | class ParserStateOp;
      |       ^
3082 warnings and 9 errors generated.
ninja: build stopped: subcommand failed.
[1]    2813959 exit 1     ./build_tools/build_p4c_with_p4mlir_ext.sh
```

lib/Transforms/SimplifyParsers.cpp
xl4624: I ended up putting it under a method of `ParserOp` because we pass in the parser's symbol table and this seemed a bit more intuitive to me

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,168 @@
...
+    void buildIndegreeAndSuccessor(
+        P4HIR::ParserOp &parser, mlir::SymbolTable &symbolTable,
+        llvm::DenseMap<P4HIR::ParserStateOp, unsigned> &indegree,
+        llvm::DenseMap<P4HIR::ParserStateOp, P4HIR::ParserStateOp> &succ);
+};
+}  // end anonymous namespace
+
+llvm::DenseSet<P4HIR::ParserStateOp> SimplifyParsers::findReachableStates(
+    P4HIR::ParserOp &parser, mlir::SymbolTable &symbolTable) {
+    auto start = symbolTable.lookup<P4HIR::ParserStateOp>("start");
```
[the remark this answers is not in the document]
xl4624: Got the start state using the parser's transition.

lib/Transforms/SimplifyParsers.cpp
xl4624: Removed any special handling, now start/accept/reject can all be collapsed.

lib/Transforms/SimplifyParsers.cpp
xl4624: Wasn't sure where to implement GraphTraits so I just left it as a parser method for now.

xl4624:
Accidentally deleted all my p4mlir git branches :skull: thankfully GitHub saved them

include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
```diff
@@ -49,6 +52,10 @@ def ParserStateOp : P4HIR_Op<"state",
  let assemblyFormat = [{
    $sym_name (`annotations` $annotations^)? $body attr-dict
  }];
+
+ let extraClassDeclaration = [{
+    mlir::Operation* getTerminator() { return getBody().back().getTerminator(); }
```
asl, line 57: I would probably not spell here as "terminator", but rather something around transition?

lib/Transforms/SimplifyParsers.cpp
asl: But we can grab the parent op from `ParserStateOp`.

include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
```diff
@@ -38,6 +38,9 @@ def ParserOp : P4HIR_Op<"parser",
     }
 
     void createEntryBlock();
+
+    llvm::SmallVector<mlir::Operation *> getSuccessors(P4HIR::ParserStateOp &state,
+                                                          mlir::SymbolTable &symbolTable);
```
asl, line 43: Why do we need to pass symboltable from outside?

include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
xl4624: changed to `getNextTransition()`

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+    llvm::SmallVector<P4HIR::ParserStateOp> worklist{start};
+    llvm::DenseSet<P4HIR::ParserStateOp> visited{start};
+
+    bool acceptReachable = false, rejectReachable = false;
+
+    while (!worklist.empty()) {
+        auto state = worklist.pop_back_val();
+        for (auto nextOp : state.getSuccessors()) {
+            auto next = llvm::cast<P4HIR::ParserStateOp>(nextOp);
+            if (mlir::isa<P4HIR::ParserAcceptOp>(next.getNextTransition())) {
```
asl, line 41: We now have `isAccept` and `isReject` for states.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+    P4HIR::ParserOp &parser, mlir::SymbolTable &symbolTable) {
+    auto start = parser.getStartState();
+    llvm::SmallVector<P4HIR::ParserStateOp> worklist{start};
+    llvm::DenseSet<P4HIR::ParserStateOp> visited{start};
+
+    bool acceptReachable = false, rejectReachable = false;
+
+    while (!worklist.empty()) {
+        auto state = worklist.pop_back_val();
+        for (auto nextOp : state.getSuccessors()) {
```
asl, line 39: it is `getNextStates` now.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "p4mlir/Transforms/Passes.cpp.inc"
+
+namespace {
+struct SimplifyParsers : public impl::SimplifyParsersBase<SimplifyParsers> {
+    void runOnOperation() override;
+
+ private:
+    /// Finds all states reachable from the 'start' state.
+    llvm::DenseSet<P4HIR::ParserStateOp> findReachableStates(P4HIR::ParserOp &parser,
+                                                             mlir::SymbolTable &symbolTable);
```
asl, line 22: `symbolTable` is unused now. Pass `ParserOp` by value here (it is just a wrapper over pointer).

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+            // Splice all operations from 'next' into the head body.
+            headBlock.getOperations().splice(headBlock.end(), nextBlock.getOperations());
+            next.erase();
+        }
+    }
+}
+
+void SimplifyParsers::runOnOperation() {
+    getOperation()->walk([&](P4HIR::ParserOp parser) {
+        mlir::SymbolTable symbolTable(parser);
```
asl, line 102: `symbolTable` is unused now. See above about passing `ParserOp` by value

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+void SimplifyParsers::collapseChains(P4HIR::ParserOp &parser, mlir::SymbolTable &symbolTable) {
+    mlir::DenseMap<P4HIR::ParserStateOp, unsigned> indegree;
+    // succ[s1] = s2 if there is exactly one outgoing edge from s1 to s2.
+    mlir::DenseMap<P4HIR::ParserStateOp, P4HIR::ParserStateOp> succ;
+
+    // We diconnect any annotated states since they can't be collapsed
+    // and if we kept them they'd stop any merging downstream.
+    for (auto state : parser.getBody().getOps<P4HIR::ParserStateOp>()) {
```
asl, line 63:
```suggestion
    for (auto state : parser.states()) {
```

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+        if (state.getAnnotations()) continue;
+
+        llvm::SmallVector<mlir::Operation *> successors = state.getSuccessors();
+        for (auto nextOp : successors) {
+            auto next = llvm::cast<P4HIR::ParserStateOp>(nextOp);
```
asl, line 68: And with `getSuccessors` you do not need these casts.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+}
+
+void SimplifyParsers::runOnOperation() {
+    getOperation()->walk([&](P4HIR::ParserOp parser) {
+        mlir::SymbolTable symbolTable(parser);
+
+        llvm::DenseSet<P4HIR::ParserStateOp> reachable = findReachableStates(parser, symbolTable);
+        for (auto s : llvm::make_early_inc_range(parser.getBody().getOps<P4HIR::ParserStateOp>()))
```
asl, line 105:
```suggestion
        for (auto s : llvm::make_early_inc_range(parser.states()))
```

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+    while (!worklist.empty()) {
+        auto state = worklist.pop_back_val();
+        for (auto nextOp : state.getSuccessors()) {
+            auto next = llvm::cast<P4HIR::ParserStateOp>(nextOp);
+            if (mlir::isa<P4HIR::ParserAcceptOp>(next.getNextTransition())) {
+                acceptReachable = true;
+            } else if (mlir::isa<P4HIR::ParserRejectOp>(next.getNextTransition())) {
+                rejectReachable = true;
+            }
+            if (visited.insert(next).second) worklist.push_back(next);
```
asl, line 46: Technically now with graph traits we could've used LLVM's `DepthFirstIterator` to iterate over all reachable states and hide the iteration boilerplate. But then we'd need to check visited set for `accept` / `reject` additionally.

I do not have strong preference for the approach, only if you'd want to try these traits.

build_tools/build_p4c_with_p4mlir_ext.sh
```diff
@@ -35,6 +35,7 @@ CMAKE_FLAGS+=" -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER:-clang++}"
 
 # Configure P4C CMake flags
 # https://github.com/p4lang/p4c/blob/main/CMakeLists.txt
+CMAKE_FLAGS+=" -DCMAKE_BUILD_TYPE=DEBUG"
```
asl, line 38: why do we need this?

xl4624: Oh oops I was doing some merge conflicts so the remote branch is messy right now (I just pass this whenever I want to run `gdb`)

lib/Transforms/SimplifyParsers.cpp
xl4624: I'll read into this (and fix the error on `next.getAnnotations()`) when I get the chance tomorrow. Thank you so much for the help!

build_tools/build_p4c_with_p4mlir_ext.sh
asl: ok, no worries. Looks like we're having test crashes now. So, it might be some iterators are wrong / messy.

asl: And finally, it seems that you're iterating over empty `succ` :)

xl4624: Ok! I think my latest commit fixed it. Here's also what the debug logs look like:

```bash
--- Simplifying parser 'p' ---
DFS visiting start
DFS visiting s0
DFS visiting s1
DFS visiting accept
DFS visiting reject
Chain starting with start
        Collapsing s0
        Collapsing s1

--- Simplifying parser 'branching' ---
DFS visiting start
DFS visiting s0
DFS visiting s1
DFS visiting s3
DFS visiting s4
DFS visiting accept
DFS visiting s2
Chain starting with s3
        Collapsing s4
        Collapsing accept
Chain starting with start
        Collapsing s0

--- Simplifying parser 'annotations' ---
DFS visiting start
DFS visiting s0
DFS visiting s1
DFS visiting accept
Chain starting with s1
        Collapsing accept
```

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+void SimplifyParsers::runOnOperation() {
+    getOperation()->walk([&](P4HIR::ParserOp parser) {
+        LLVM_DEBUG(llvm::dbgs() << "\n--- Simplifying parser '" << parser.getName() << "' ---\n");
+        llvm::df_iterator_default_set<llvm::GraphTraits<P4HIR::ParserOp>::NodeRef> reachable;
+        bool acceptReachable = false, rejectReachable = false;
+
+        /// Finds all states reachable from the start state and deletes unreachable states.
+        for (auto it = llvm::df_ext_begin(parser, reachable);
```
asl, line 83: Can't you just use `for (auto op: depth_first_ext(parser, reachable))` here?

test/Transforms/Passes/simplify-parsers.mlir
```diff
@@ -0,0 +1,144 @@
...
+// CHECK-NOT:   p4hir.state @accept
+// CHECK-NOT:   p4hir.state @reject
+module {
+  p4hir.parser @branching()() {
+    p4hir.state @start {
+      p4hir.transition to @branching::@s0
+    }
+    p4hir.state @s0 {
+      %true = p4hir.const #true
+      p4hir.transition_select %true : !p4hir.bool {
```
asl, line 84: Probably would make sense to have another testcase with `transition_select` having a single case.

test/Transforms/Passes/simplify-parsers.mlir
```diff
@@ -0,0 +1,144 @@
...
+// CHECK:       p4hir.parser_accept
+
+// CHECK-NOT:   p4hir.state @accept
+// CHECK-NOT:   p4hir.state @reject
+module {
+  p4hir.parser @annotations()() {
+    p4hir.state @start {
+      p4hir.transition to @annotations::@s0
+    }
+    p4hir.state @s0 annotations {name = "state.s0"} {
```
asl, line 130: Can we have testcase similar to branching, but with one of destination set having annotations?

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,103 @@
...
+             it != llvm::df_ext_end(parser, reachable); ++it) {
+            auto state = mlir::cast<P4HIR::ParserStateOp>(*it);
+            LLVM_DEBUG(llvm::dbgs() << "DFS visiting " << state.getName() << "\n");
+            if (state.isAccept()) acceptReachable = true;
+            if (state.isReject()) rejectReachable = true;
+        }
+        if (!acceptReachable && !rejectReachable)
+            parser.emitWarning("Parser never reaches the 'accept' or 'reject' state.");
```
asl, line 91: In the original pass, this is an error.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+        for (auto state : llvm::make_early_inc_range(parser.states()))
+            if (!reachable.contains(state)) state.erase();
```
asl, line 94: I would add diagnostics from the original pass. Both error / warnings as well as debug printing.

xl4624: Added the error/warnings + debug printing from p4c's pass, also mirrored their behavior of not removing the accept state if it's not reachable

lib/Transforms/SimplifyParsers.cpp
xl4624: changed to `parser.emitError()`

test/Transforms/Passes/simplify-parsers.mlir
xl4624: Added `@annotated_branching` which includes a single case `transition_select` as well. This test case made me realize that my original way of finding chain heads was just incorrect (I was using `if (indegree[state] == 1) continue; // will get merged by another chain`). When 2 states transition to a single state and one of them is annotated, since we skip annotated states when calculating indegrees then that single state has an indegree of 1 and would never get merged even if it can. Tried a couple more approaches in how to skip annotations "correctly" and then just gave up and just copied the p4c approach of finding chain heads 🙃

lib/Transforms/SimplifyParsers.cpp
asl: Not sure if this is something desired or not, but for now let's be bug-per-bug compatible :)

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    // We diconnect any annotated states since they can't be collapsed
+    // and if we kept them they'd poison any merging downstream.
+    for (auto state : parser.states()) {
+        if (outdegree[state] != 1 || state.getAnnotations()) continue;
+        P4HIR::ParserStateOp successor = *state.getNextStates().begin();
+        if (indegree[successor] != 1 || successor.getAnnotations()) continue;
+
+        succ[state] = successor;
+        pred[successor] = state;
```
asl, line 51: Can we simplify here a bit? We just need heads here, right? Can we just check if their outdegree is 1 (plus usual annotations checks) and indegree != 1. And make a worklist of heads? Then you'd just iterate below over the worklist.

lib/Transforms/SimplifyParsers.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    // succ[s1] = s2 if there is exactly one outgoing edge from s1 to s2
+    // and s2 has exactly one incoming edge from s1.
+    mlir::DenseMap<P4HIR::ParserStateOp, P4HIR::ParserStateOp> succ;
+    mlir::DenseMap<P4HIR::ParserStateOp, P4HIR::ParserStateOp> pred;
+
+    // We diconnect any annotated states since they can't be collapsed
+    // and if we kept them they'd poison any merging downstream.
+    for (auto state : parser.states()) {
+        if (std::distance(state.getNextStates().begin(), state.getNextStates().end()) != 1)
```
asl, line 44:
```suggestion
        if (!llvm::hasNItems(state.getNextStates(), 1))
```

xl4624:
Yea sorry I was rebuilding from source so I was using the CI/CD to run tests while trying out your approach. If I understood you properly on what to check, you're saying that we could add states that have an outdegree of 1 and indegree != 1 as heads. I think the issue here is with annotations, and that we would need 2 bits of information, let's say on 2 small examples:

Example 1:
```
s0 (annotated) -> s2
s1 -> s2
s2 -> s3
```

and

Example 2:
```
s1 (annotated) -> s2 -> s3
```

For both examples here we would like to collapse using s2 as the head. But if we were checking for total indegree != 1, in Example 2 we would skip s2 as a head, even though it should be a head, because the state that has an edge to s2 is annotated and can't be a head.
If we only checked for non-annotated indegree, then Example 1 has an issue, since we can't start a chain at s1 (the non-annotated edge pointing to s2) as there are multiple states with an edge pointing to s2.

The condition for any state would have to be that the total indegree != 1 and annotated indegree != 1, I think? We could go with the approach of finding both types of indegrees for each state, or if there is an even simpler approach you had in mind. Sorry it's kind of a rushed explanation as I have an exam in an hour 😅

asl:
There is another weird case: parser graphs might have loops. So we need to ensure this also works properly.

Also, there is special case for `@name` annotation: if it is the only annotation we can merge *into* this state (so it could be head despite having the annotation).

So yes, after second thought looks like we need to keep track of both predecessors and successors. Will you just handle two cases above? For cycles we'd likely only need a testcase?

xl4624:
Added a test case for cycles, which just replicates how `p4c` does it and leaves cycles unmerged. You could technically collapse simple cycles into a single state that self-transitions, but I'd imagine that would have to have to use a strongly connected components algorithm like the one in `p4c`s `CallGraph` since you can't just check if the state is in `pred`.

Also added the special case for `@name` annotations: the `@annotations` test case now has an `@atomic` state that doesn't do any collapsing, and a `@name` state that can act as a head (but doesn't get added into another chain).

Thanks again for the review and comments @asl

asl (approved):
Thanks! Let's have this as a first approximation. We might refine it later!

PR #167, opened 2025-05-01
Add some convenient parser IR methods

asl:
Also add graph traits and ensure we can use them

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
@@ -41,8 +41,10 @@ def ParserOp : P4HIR_Op<"parser",
 
     /// IR-related methods
 
-    /// Returns the iterator over all states defined in the parswr
-    auto states() { return getBody().getOps<P4HIR::ParserStateOp>(); }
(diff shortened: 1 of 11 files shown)


asl:
@xl4624 Please take a look. This adds some parser IR methods, plus graph traits and some example use. I think it would be great if your code for parser simplification would use it.

include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
```diff
@@ -37,18 +39,64 @@ def ParserOp : P4HIR_Op<"parser",
       return {};
     }
 
+    /// IR-related methods
+
+    /// Returns the range over all states defined in the parswr
```
qobilidop, line 44: Nit: typo "parswr" -> "parser"

qobilidop (approved):

include/p4mlir/Dialect/P4HIR/P4HIR_ParserOps.td
asl, line 44:
```suggestion
    /// Returns the range over all states defined in the parser
```

Issue #139, opened 2025-03-24
Create a BMv2 dialect prototype


qobilidop:
This is an overall tracking issue to create a BMv2 dialect in P4MLIR.

[BMv2](https://github.com/p4lang/behavioral-model) is a popular software simulator target for P4. In our current open source P4 compiler [P4C](https://github.com/p4lang/p4c), when targeting BMv2, a P4 program is converted to a JSON file, which BMv2 uses as a specification for processing packets. In P4MLIR, we plan to add a dialect specifically for modeling [BMv2 JSON primitives](https://github.com/p4lang/behavioral-model/blob/main/docs/JSON_format.md), so that the BMv2 dialect -> BMv2 JSON transformation could be straightforward.

In the longer term, we expect a compilation path like P4C frontend -> P4HIR dialect -> BMv2 dialect -> BMv2 JSON. In this prototype, we will concentrate on implementing a subset of BMv2 JSON primitives in the BMv2 dialect, and implementing the corresponding BMv2 dialect -> BMv2 JSON transformation. The scope of the subset is left for follow-up discussion.

asl:
@qobilidop I remember BMv2 has some restrictions on kinds of expressions it does support (I do not remember offhand). So likely these are to be lowered before translation to dialect, right?

qobilidop:
Yes, BMv2 definitely has such restrictions. Not all valid P4 programs are also valid BMv2 programs. https://github.com/p4lang/project-ideas/issues/8 and https://github.com/p4lang/project-ideas/issues/33 are some examples I'm aware of, and there are likely more.

I guess when lowering from P4HIR to BMv2 dialect, we could do some checks to reject invalid programs upfront. Current P4C BMv2 backend already has such checks like https://github.com/p4lang/p4c/blob/03c6717fd3e6bd2db969eab8ff0e4553dd8aa26e/backends/bmv2/common/check_unsupported.h#L27-L36.

asl:
Ah, loops, right. But I thought about something more basic... Do not recall offhand, but something like simple conditions or so. Anyway, this will be uncovered sooner than later :)

Joejiong:
cause when I look at the spec, bmv2 is a target, and above that is architecture, when I surveyed my previous task '@optional', I also noticed there are design choices of how to model bmv2, previously I thought is bmv2 first then underlying v1/psa/pna, but when I check the spec, it seems that this is more logic, I don't know if I'm on the right track. @qobilidop 
[...]

PR #168, opened 2025-05-04
[WiP] Add an P4HIR-to-P4 syntax exporter.

fruffy:
**Not ready yet for review. Currently intended to check CI.** 

This PR adds a utility to export P4HIR generated by the translate pass back to P4. The purpose is threefold. First, check that the current P4HIR representation is semantically equivalent to the original P4 source program and retains all the information needed for a valid P4 program. Second, provide a utility similar to ToP4, which can export P4 code for manual inspection and as reference. Third, once more front ends are available, serve as a transpiler back end to convert a particular source language (C/Python...) to P4. 

The current approach to the exporter is as follows. 
1. First, collect all complex types (headers, enums, errors) in the program and declare them. 
2. Iterate over all top-level ops (externs, functions, actions, controls, parsers, packages) in the module and declare/define them as needed.

For each block that corresponds to a P4 block we hoist all variable declarations. All other Ops are handled in order.

We currently use two checks to ensure that this pass works. First, we use references files for each test P4 program and we compare against these references. That also makes reviewing changes easier. Second, we feed the generated program to p4test to make sure it compiles. This ensures that we always generate a valid P4 program with this exporter.

Tasks left. 

- [ ] Figure out a way to correctly declare controls and parsers. Currently we are missing type and direction information for parameters. As long as this feature is missing we can not export real programs.
- [ ] Method calls are currently scope operators with several operators. This can be difficult to convert back into a method call. In particular, when these are part of a table action list. 
- [ ] Comments are all over the place. Needs to be cleaned up.
- [ ] General clean up (smaller function, remove unnecessary code, etc).
- [ ] Clean up error reporting.
- [ ] Biggest TODO: Make sure all current example programs pass.


qobilidop:
> The current approach to the exporter is as follows.
>
> 1. First, collect all complex types (headers, enums, errors) in the program and declare them.
> 2. Iterate over all top-level ops (externs, functions, actions, controls, parsers, packages) in the module and declare/define them as needed.

The general approach I planned to adopt for #139 is similar! The top-level structure of the [BMv2 JSON format](https://github.com/p4lang/behavioral-model/blob/main/docs/JSON_format.md) is essentially these top-level types and ops listed in categories. I'll likely take some inspirations from your work here :)

PR #175, opened 2025-05-15
Update P4C reference and add some linter configurations. Compile with DebugInfo.

fruffy:
Recent P4C updates include some cpplint updates which can also be useful here. Also add some more configurations specific to this repository. The `.clang-tidy` file could also be a symlink.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+CompileFlags:
+  CompilationDatabase: third_party/p4c/build/
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,27 @@
+---
+Checks:
+'clang-diagnostic-*,
+clang-analyzer-*,
+modernize-*,
+performance-*,
+readability-*,
+bugprone-*,
+cppcoreguidelines-*,
+boost-*,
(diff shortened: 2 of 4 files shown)


.clangd
```diff
@@ -0,0 +1,2 @@
+CompileFlags:
+  CompilationDatabase: third_party/p4c/build/
```
asl, line 2: This tied to the way how the things are built on CI, but effectively disallows e.g. just to drop stuff into `extensions` of existing p4c. So I'd rather not include here as this depends on the particular build directory layout.

fruffy: `third_part/p4c` only exists for the purpose of CI? I assumed it is meant to be a standalone build.

asl: It works just fine compiling as a standalone extension, so I would explicitly not to disallow this possibility (and actually this is how I'm compiling the things locally).

fruffy: Iirc the compilation database here is just a hint, you could also pass it to your IDE options of your choice. But this gives a nicer out-of-the-box experience. Let me check whether multiple folders are possible.

asl (approved):

PR #176, opened 2025-05-16
Allow nameless (anonymous) enums

### LINKED ISSUE #174: Action Enum should be a special enum type
`Type_ActionEnum` is converted into a enum type with name "action_enum". This could create ambiguities if an enum is named `action_enum`.

asl:
These are used to represent action enums, where we really do not care about name, but only need the list of fields (also preventing clashes with user-provided enums in theory).

Fixes #174

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
@@ -580,9 +580,21 @@ def EnumType : P4HIR_Type<"Enum", "enum", []> {
                    annotations && !annotations.empty() ? annotations : mlir::DictionaryAttr());
     }]>,
 
+     TypeBuilder<(ins "llvm::ArrayRef<mlir::Attribute>":$fields,
+                     CArg<"mlir::DictionaryAttr", "{}">:$annotations), [{
+      return $_get($_ctxt, "", mlir::ArrayAttr::get($_ctxt, fields),
+                   annotations && !annotations.empty() ? annotations : mlir::DictionaryAttr());
+    }]>,
+
     TypeBuilder<(ins "llvm::StringRef":$name, "mlir::ArrayAttr":$fields,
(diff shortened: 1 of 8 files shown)


fruffy (approved):
Thanks! Basically for any enum without a name we can treat them as anonymous. What about struct types? E.g., what tables return.

asl:
> Thanks! Basically for any enum without a name we can treat them as anonymous. What about struct types? E.g., what tables return.

Yes, this will follow :)

asl:
> What about struct types? E.g., what tables return.

Actually, these are represented via normal structs in P4C IR. With the name equal to the name of the table:
```
            type: [2030] Type_Struct name=t declid=20
              typeParameters: [2032] TypeParameters
              fields: [2021] StructField name=hit declid=65
                type: [551] Type_Boolean
              fields: [2023] StructField name=miss declid=66
                type: [551] Type_Boolean
              fields: [2026] StructField name=action_run declid=67
                type: [2025] Type_ActionEnum...
```

Do we want to make them special? Probably not as there is no way we can create type clash with something user as there is `action_run` field that is not-representable via user-written P4.

fruffy:
> Actually, these are represented via normal structs in P4C IR. With the name equal to the name of the table:

I think this is a small oversight, since these types are generated ad-hoc: https://github.com/p4lang/p4c/blob/6f6048a1daf14ffadeb56b0dad843b8cd7390464/ir/ir.cpp#L193

>Do we want to make them special? Probably not as there is no way we can create type clash with something user as there is action_run field that is not-representable via user-written P4.

From my point of view, associating them with a trait that they are "anonymous"/"local" is sufficient.

asl:
> I think this is a small oversight, since these types are generated ad-hoc: https://github.com/p4lang/p4c/blob/6f6048a1daf14ffadeb56b0dad843b8cd7390464/ir/ir.cpp#L193

I guess the main question is: what if we'd have empty name at P4C side here? Will something break? :)

PR #194, opened 2025-06-10
Add RemoveControlFlowPass

elhewaty:


xl4624:
Another thing to watch out for is that in P4C this pass requires all declarations to be at the "top", basically moving them all before the first state, with `moveDeclaration` pass (which itself requires `uniqueNames`)

I looked into this before and ran into issues with something like (dumb example):

```
parser p() {
	state start {
		bool b = true;
		if (b) {
			b = false;
		}
		statement;
		use b;
  }
}
```

But if you ran this example without any processing you'd get:
```
parser p() {
	state start {
		bool b = true;
		transition select (b) {
			true : start_true;
			_ : start_join;
		}
	}
	state start_true {
		b = false; // b is not defined here (same in mlir)
	}
	state start_join {
		statement;
		use b; // again can't use b
	}
}
```

Good luck!

asl:
> Another thing to watch out for is that in P4C this pass requires all declarations to be at the "top", basically moving them all before the first state, with `moveDeclaration` pass

This is just the limitation of P4C pass (as there is no sane way to track normal def-use chains there). We do not need *all* declarations to be moved there, only those variable that are defined in the "top" part of the state and used in the "bottom" one.

elhewaty:
> This is just the limitation of P4C pass (as there is no sane way to track normal def-use chains there). We do not need _all_ declarations to be moved there, only those variables that are defined in the "top" part of the state and used in the "bottom" one.

@asl I am thinking of implementing two passes for renaming and hoisting vars, what do you think?

elhewaty:
> > don't you think we need to implement the hoisting pass for similar cases mentioned here [#194 (comment)](https://github.com/p4lang/p4mlir-incubator/pull/194#issuecomment-2957664593) if not what do you suggest?
> 
> Why this is needed? What problems will it solve? This is exactly why you'd need to start from RFC.

I am not sure if I am correct, but after some thinking, MLIR is an SSA form so this thing is impossible to find in mlir

```
parser p() {
	state start {
		bool b = true;
		if (b) {
			b = false;
		}
		statement;
		use b;
  }
}
```

it will be lowered like this -I think-

```
#false = #p4hir.bool<false> : !p4hir.bool
#true = #p4hir.bool<true> : !p4hir.bool
module {
  p4hir.parser @simple_parser(%arg0: !p4hir.ref<!p4hir.bool>, %arg1: !p4hir.ref<!p4hir.bool>)() {
    p4hir.state @start {
      %false = p4hir.const #false
      p4hir.assign %false, %arg1 : <!p4hir.bool>
      %condition = p4hir.read %arg1 : <!p4hir.bool>
      p4hir.if %condition {
        %true = p4hir.const #true
        p4hir.assign %true, %arg1 : <!p4hir.bool>
      }
      p4hir.transition to @simple_parser::@final_state
    }
    p4hir.state @final_state {
      // differenct scope, so different condition
      %condition = p4hir.read %arg1 : <!p4hir.bool>
      p4hir.assign %condition, %arg0 : <!p4hir.bool>
      p4hir.parser_accept
    }
    p4hir.transition to @simple_parser::@start
  }
}
``` 

so, converting if into select will not need all dummy complex things I suggested (if this is not dummy, too).
@asl what do you think?

elhewaty:
I think I don't need the renaming, and hoisting pass, as it is SSA form after all, so all vars in the same states are unique. but we need to move only Declarations in the top of the state that have uses after the `IfOp`, and also we need to create new states for the `Then` and `Else` regions. Given that a state can have multiple `IfOp`s, so we need to rename the newly created states, I am thinking of adding a number as a suffix to the state name and a true/false string part
```
state a {}
state a_true_1 {}
state a_false_1 {}
state a_true_2 {}
state a_false_2 {}
```

regarding the variable names: as all states in the same scope are unique so for the state vars

`var name = state_name + var_name`

```
// global scope
var a_x = ...
state a {}

```

@asl what do you think?

elhewaty:
@asl, It doesn't handle switch yet. but I thought it is a good idea to show you where did I reach.

input
```
#false = #p4hir.bool<false> : !p4hir.bool
#true = #p4hir.bool<true> : !p4hir.bool
module {
  p4hir.parser @simple_parser(%arg0: !p4hir.ref<!p4hir.bool>, %arg1: !p4hir.ref<!p4hir.bool>)() {
    p4hir.state @start {
      %false = p4hir.const #false
      p4hir.assign %false, %arg1 : <!p4hir.bool>
      %condition = p4hir.read %arg1 : <!p4hir.bool>
      %bla = p4hir.read %arg0 : <!p4hir.bool>
      p4hir.if %condition {
        p4hir.assign %condition, %arg1 : <!p4hir.bool>
      } else {
        p4hir.assign %bla, %arg0 : <!p4hir.bool>
      }
      p4hir.transition to @simple_parser::@final_state
    }
    p4hir.state @final_state {
      // different scope, so different condition 
      %condition = p4hir.read %arg1 : <!p4hir.bool>
      p4hir.assign %condition, %arg0 : <!p4hir.bool>
      p4hir.parser_accept
    }
    p4hir.transition to @simple_parser::@start
  }
}
```

output
```
#false = #p4hir.bool<false> : !p4hir.bool
#true = #p4hir.bool<true> : !p4hir.bool
module {
  %true = p4hir.const #true
  %false = p4hir.const #false
  p4hir.parser @simple_parser(%arg0: !p4hir.ref<!p4hir.bool>, %arg1: !p4hir.ref<!p4hir.bool>)() {
    %val = p4hir.read %arg0 : <!p4hir.bool>
    %val_0 = p4hir.read %arg1 : <!p4hir.bool>
    p4hir.state @start {
      p4hir.assign %false, %arg1 : <!p4hir.bool>
      p4hir.transition_select %val_0 : !p4hir.bool {
        p4hir.select_case {
          p4hir.yield %true : !p4hir.bool
        } to @simple_parser::@start_if_true_0
        p4hir.select_case {
          p4hir.yield %true : !p4hir.bool
        } to @simple_parser::@start_if_true_0
      }
    }
    p4hir.state @start_if_false_0 annotations {} {
      p4hir.assign %val, %arg0 : <!p4hir.bool>
      p4hir.transition to @simple_parser::@start_if_join_0
    }
    p4hir.state @start_if_true_0 annotations {} {
      p4hir.assign %val_0, %arg1 : <!p4hir.bool>
      p4hir.transition to @simple_parser::@start_if_join_0
    }
    p4hir.state @start_if_join_0 annotations {} {
      p4hir.transition to @simple_parser::@final_state
    }
    p4hir.state @final_state {
      %val_1 = p4hir.read %arg1 : <!p4hir.bool>
      p4hir.assign %val_1, %arg0 : <!p4hir.bool>
      p4hir.parser_accept
    }
    p4hir.transition to @simple_parser::@start
  }
}
```

asl:
```mlir
        p4hir.select_case {
          p4hir.yield %true : !p4hir.bool
        } to @simple_parser::@start_if_true_0
```

Select cases are expected to yield sets. We have TBD for the verifier here.

So, it should be:
```mlir
        p4hir.select_case {
          %set = p4hir.set (%true) : !p4hir.set<!p4hir.bool>
          p4hir.yield %set : !p4hir.set<!p4hir.bool> 
        } to @simple_parser::@start_if_true_0
```

PR #195, opened 2025-06-12
Make actions / functions / controls isolated

asl:
Ensure that:
 - Constants are materialized when used
 - Control locals (including arguments) are available for nested actions via symbols
 - Lots of other fixes here and there

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_ControlOps.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_ControlOps.td
@@ -3,7 +3,8 @@
 
 def ControlOp : P4HIR_Op<"control",
     [Symbol, SymbolTable, Annotated,
-     FunctionOpInterface, AutomaticAllocationScope]> {
+     FunctionOpInterface, AutomaticAllocationScope,
+     IsolatedFromAbove]> {
   let arguments = (ins SymbolNameAttr:$sym_name,
                        TypeAttrOf<FuncType>:$applyType,
                        TypeAttrOf<CtorType>:$ctorType,
--- a/tools/p4mlir-translate/translate.cpp
+++ b/tools/p4mlir-translate/translate.cpp
@@ -2372,7 +2372,7 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
             switch (auto dir = param->direction) {
                 case P4::IR::Direction::None:
                 case P4::IR::Direction::In: {
-                    auto paramType = getOrCreateType(param->type);
+                    auto paramType = getOrCreateType(param);
 
                     // Nothing to do special, just pass things direct
                     visit(arg->expression);
@@ -2434,8 +2434,8 @@ bool P4HIRConverter::preorder(const P4::IR::MethodCallExpression *mce) {
                 // Grab a suitable placeholder for it.
                 mlir::Value placeholder;
                 if (param->isOptional()) {
-                    placeholder = builder.create<P4HIR::UninitializedOp>(
-                        getLoc(builder, mce), getOrCreateType(param->type));
+                    placeholder = builder.create<P4HIR::UninitializedOp>(getLoc(builder, mce),
+                                                                         getOrCreateType(param));
                 } else {
                     BUG_CHECK(param->direction == P4::IR::Direction::None,
                               "control plane values should be directionless");
@@ -2631,7 +2631,7 @@ bool P4HIRConverter::preorder(const P4::IR::ConstructorCallExpression *cce) {
             // Create a placeholder for @optional arguments
             if (!argVal && param->isOptional())
                 argVal = builder.create<P4HIR::UninitializedOp>(getLoc(builder, cce),
-                                                                getOrCreateType(param->type));
+                                                                getOrCreateType(param));
             BUG_CHECK(argVal, "unconverted argument for parameter %1%", param);
             operands.push_back(argVal);
         }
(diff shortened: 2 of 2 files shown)


asl:
Ok, so this is quite large and invasive IR change that ensures that nested actions are properly isolated from outer code and therefore we can process all of them independently.

This required some decent amount of changes, but the core of them is straightforward: we use standard symbol mechanism to access values "from outside world". Normally we do this for calls (normal and method calls) and parser states. Here this is extended to:
  - Control locals (including control arguments)
  - Instantiations

This also required splitting some ops and providing couple of new ones:
 - We separate instantiations from constructions (in terms of P4C AST IR: `Declaration_Instance` and `ConstructorCallExpression`. The former do produce symbols, the latter – don't.
 - We're having ops to reference symbols (and produce normal SSA value) and typing locals to symbols

Some checks are refined here and there and become more clear. The IR changes could be seen in test/Dialect/P4HIR/control.mlir

fruffy:
> My understanding from last week's meeting is that @fruffy is planning to give this PR a proper review by trying it out with some edge cases. Is that still the plan?

Most of the edge cases are part of the compiler tests already I believe. But we would need to check whether they are handled correctly. Cases such as

```
control ingress(inout Parsed_packet h, inout Metadata m, inout standard_metadata_t sm) {
    action do_action(inout bit<8> val_1, inout bit<8> val_2) {
        val_1 = 8w1;
        val_2 = 8w2;
        h.h.b = 8w3;
    }
    apply {
        do_action(h.h.b, h.h.b);
    }
}
```
are technically valid.

asl:
> > My understanding from last week's meeting is that @fruffy is planning to give this PR a proper review by trying it out with some edge cases. Is that still the plan?
> 
> Most of the edge cases are part of the compiler tests already I believe. But we would need to check whether they are handled correctly. Cases such as

This is already supported as we model copy-in / copy-out explicitly. For example, for the code:
```p4
extern Crc16 {
    int<16> hash<U>(in U input_data);
    U id<U>(in U x);
    Crc16(int a);
}

control InnerPipe(bit<10> arg1, in int<16> arg2, out int<16> arg3)(bool flag) {
    apply {}
}

control Pipe1(bit<10> arg1, in int<16> arg2, out bit<10> oarg2) {
    const int topC = 42;
    bit<10> topV = 10;

    InnerPipe(true) inner1;
    InnerPipe(false) inner2;
    Crc16(42) crc16;

    action foo(inout bit<10> x2) {
        int<16> x1 = 3 + arg2 + topC;
        x1 = 4 + crc16.hash(x1);
        x2 = x2 + 42;
        topV = topV + x2;
    }
    action bar() {
        bit<10> x1 = 2 + topV;
        x1 = x1 - arg1;
        oarg2 = x1;
        topV = x1;
    }
    apply {
        bit<10> x1 = arg1;
        int<16> x2 = 5;
        x2 = arg2;
        bar();
        if (arg2 == 3) {
            foo(topV);
            x2 = 3;
        }
    }
}
```

we are generating:
```mlir
...
    %c10_b10i = p4hir.const #int10_b10i
    %cast = p4hir.cast(%c10_b10i : !b10i) : !b10i
    %topV = p4hir.variable ["topV", init] : <!b10i>
    p4hir.assign %cast, %topV : <!b10i>
    p4hir.control_local @__local_Pipe1_topV_0 = %topV : !p4hir.ref<!b10i>
...
    p4hir.func action @foo(%arg3: !p4hir.ref<!b10i> {p4hir.dir = #p4hir<dir inout>, p4hir.param_name = "x2"}) {
      %c3_i16i = p4hir.const #int3_i16i
      %__local_Pipe1_arg2_0 = p4hir.symbol_ref @Pipe1::@__local_Pipe1_arg2_0 : !i16i
      %add = p4hir.binop(add, %c3_i16i, %__local_Pipe1_arg2_0) : !i16i
      %c42_i16i = p4hir.const #int42_i16i
      %add_0 = p4hir.binop(add, %add, %c42_i16i) : !i16i
      %x1 = p4hir.variable ["x1", init] : <!i16i>
      p4hir.assign %add_0, %x1 : <!i16i>
      %c4_i16i = p4hir.const #int4_i16i
      %val = p4hir.read %x1 : <!i16i>
      %0 = p4hir.call_method @Pipe1::@crc16::@hash<[!i16i]>(%val) : (!i16i) -> !i16i
      %add_1 = p4hir.binop(add, %c4_i16i, %0) : !i16i
      p4hir.assign %add_1, %x1 : <!i16i>
      %c42_b10i = p4hir.const #int42_b10i
      %val_2 = p4hir.read %arg3 : <!b10i>
      %add_3 = p4hir.binop(add, %val_2, %c42_b10i) : !b10i
      p4hir.assign %add_3, %arg3 : <!b10i>
      %__local_Pipe1_topV_0 = p4hir.symbol_ref @Pipe1::@__local_Pipe1_topV_0 : !p4hir.ref<!b10i>
      %__local_Pipe1_topV_0_4 = p4hir.symbol_ref @Pipe1::@__local_Pipe1_topV_0 : !p4hir.ref<!b10i>
      %val_5 = p4hir.read %__local_Pipe1_topV_0_4 : <!b10i>
      %val_6 = p4hir.read %arg3 : <!b10i>
      %add_7 = p4hir.binop(add, %val_5, %val_6) : !b10i
      p4hir.assign %add_7, %__local_Pipe1_topV_0 : <!b10i>
      p4hir.return
    }
...
    p4hir.control_apply {
...
      p4hir.if %eq {
        p4hir.scope {
          %x2_inout_arg = p4hir.variable ["x2_inout_arg", init] : <!b10i>
          %val = p4hir.read %topV : <!b10i>
          p4hir.assign %val, %x2_inout_arg : <!b10i>
          p4hir.call @Pipe1::@foo (%x2_inout_arg) : (!p4hir.ref<!b10i>) -> ()
          %val_3 = p4hir.read %x2_inout_arg : <!b10i>
          p4hir.assign %val_3, %topV : <!b10i>
        }
```

1. We are creating local variable to model copy-in / copy-out semantics of `foo` call in apply block, and we pass reference to this local copy
2. In action itself direct access to control locals is done through  `p4hir.symbol_ref` ops. Note that these are the only places where we can obtain aliased objects. But because we do not have any "pointer arithmetic" and because we refer symbols explicitly by name, we can just drop duplicate ops and ensure that are no explicit aliases
3. Inside `foo` we're reading from / writing to `arg3` argument and obtain the value from the copy of `topV`.
4. `foo` output value is written to `topV` after the action call.

PR #196, opened 2025-06-12
Use fully qualified symbols references for everything nested

asl:


asl:
This is a follow-up PR for https://github.com/p4lang/p4mlir-incubator/pull/195

Now we use fully-qualified symbol references. The ultimate goal is to resolve ambiguous symbol references in presence of name shadowing. P4 allows explicit top-level declaration references (like in https://github.com/p4lang/p4c/blob/main/testdata/p4tc_samples/global_action_example_01.p4#L99 for example). Previously we had no way to represent this as we always checked control scope first.

Now, everything is fully qualified, so in the example above we will be able to refer to top-level actions via `@send_nh` and to nested one via `@ingress:@send_nh`. This is done consistently for everything nested including control locals, actions, tables, etc.

This also simplified symbol lookup: we always start from topdown from top-level module.

qobilidop (approved):

PR #200, opened 2025-06-16
Add type interface to grab default values. Use it to implement InvalidHeader / InvalidHeaderUnion constants

asl:


include/p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.td
```diff
@@ -159,4 +159,21 @@ def AnnotatedType : TypeInterface<"AnnotatedType"> {
   let cppNamespace = "::P4::P4MLIR::P4HIR";
 }
 
+
+def HasDefaultValue : TypeInterface<"HasDefaultValue"> {
+  let description = [{
+    Implemented by the types having a default value
+    Definition: https://p4.org/p4-spec/docs/P4-16-working-spec.html#sec-default-values
```
qobilidop, line 166: Might be better to pin to a specific spec version?

qobilidop (approved):

include/p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.td
asl: Well, I'd rather update the implementation should the spec here would change :)

Issue #201, opened 2025-06-16
Implement enum elimination


asl:
Essentially, a ConvertEnums pass from mainline P4C. However, we'd just lower to serialized enums as common underlying representation. We will need some kind of external model, so the backend could specify the required natural underlying type.

asl:
In P4C there is a notion of "policy". Essentially what would specify from outside the type enum will be used (e.g. `bit<32>` or `bit<64>` and so on).

In MLIR we have interfaces. Overall we can have "enum representation" concept for enum type and provide an external model for this: https://mlir.llvm.org/docs/Interfaces/#external-models-for-attribute-operation-and-type-interfaces

So we can have, say, use `bit<32>` as a fallback, but target could attach their own elimination model to enum type.

asl:
Implemented in https://github.com/p4lang/p4mlir-incubator/pull/240

PR #209, opened 2025-06-20
Add enum elimination pass

xl4624:
Implements #201


include/p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.td
```diff
@@ -100,6 +100,26 @@ def FieldIDTypeInterface : TypeInterface<"FieldIDTypeInterface"> {
   let cppNamespace = "::P4::P4MLIR::P4HIR";
 }
 
+// TODO: Can we add a default implementation in TableGen?
```
xl4624, line 103: This requires `P4HIR::BitsType::get(...)` which I don't think is available here (since P4HIR_TypeInterfaces comes before P4HIR_Types)

Issue #202, opened 2025-06-16
Implement serialized enum elimination


asl:
Implement EliminateSerEnums from P4C midend passes. Should be pretty straightforward.

PR #210, opened 2025-06-30
Add SerEnumElimination pass

xl4624:
Implements #202, based off #209 (not sure how to stack PRs on a GitHub fork)

### DIFF
--- a/lib/Transforms/SerEnumElimination.cpp
+++ b/lib/Transforms/SerEnumElimination.cpp
@@ -1,0 +1,84 @@
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
+#include "mlir/Support/LLVM.h"
+#include "p4mlir/Conversion/ConversionPatterns.h"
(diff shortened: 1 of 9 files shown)


lib/Transforms/SerEnumElimination.cpp
```diff
@@ -0,0 +1,163 @@
...
+    : public P4::P4MLIR::impl::SerEnumEliminationBase<SerEnumEliminationPass> {
+    SerEnumEliminationPass() = default;
+    void runOnOperation() override;
+};
+
+class SerEnumTypeConverter : public TypeConverter {
+ public:
+    explicit SerEnumTypeConverter(MLIRContext *ctx) {
+        // Fallback for other types
+        addConversion([](Type type) -> Type { return type; });
```
asl, line 30: Do we correctly handle case of serenum in a header or struct? Anything `StructLike`?

xl4624: oo I did not handle that. I'll add it to enum elimination (I also forgot controls and parsers here I also just added to the other pass)

PR #213, opened 2025-07-04
Add pass to remove aliases

xl4624:


lib/Transforms/RemoveAliases.cpp
```diff
@@ -0,0 +1,103 @@
...
+            Type underlyingType = convertType(aliasType);
+
+            return llvm::TypeSwitch<Attribute, Attribute>(attr)
+                .Case<P4HIR::IntAttr>([&](P4HIR::IntAttr intAttr) {
+                    return P4HIR::IntAttr::get(underlyingType, intAttr.getValue());
+                })
+                .Case<P4HIR::BoolAttr>([&](P4HIR::BoolAttr boolAttr) {
+                    return P4HIR::BoolAttr::get(underlyingType, boolAttr.getValue());
+                })
+                .Case<P4HIR::ErrorCodeAttr>([&](P4HIR::ErrorCodeAttr errorAttr) {
```
asl, line 44: Are you sure all these are actually possible? Per P4 spec https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-newtype:
> Currently the types that can be created by the type keyword are restricted to one of: bit<>, int<>, bool, or types defined using type from such types.

and

> While similar to typedef, the type keyword introduces a new type which is not a synonym with the original type: values of the original type and the newly introduced type cannot be mixed in expressions.

Please ensure that the test would cover all these cases.

include/p4mlir/Dialect/P4HIR/P4HIR_Attrs.td
```diff
@@ -27,6 +27,9 @@ def P4HIR_BoolAttr : P4HIR_Attr<"Bool", "bool", [TypedAttrInterface]> {
     AttrBuilder<(ins "bool":$value), [{
       return $_get($_ctxt, BoolType::get($_ctxt), value);
     }]>,
+    AttrBuilderWithInferredContext<(ins "mlir::Type":$type, "bool":$value), [{
```
asl, line 30: This is nonsense. `BoolAttr` could only be of bool type.

lib/Transforms/RemoveAliases.cpp
```diff
@@ -0,0 +1,103 @@
...
+                    return P4HIR::ErrorCodeAttr::get(underlyingType, errorAttr.getField());
+                })
+                .Case<P4HIR::AggAttr>([&](P4HIR::AggAttr aggAttr) {
```
asl, line 47: This is handled by `P4HIRTypeConverter` automatically

lib/Transforms/RemoveAliases.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    AliasTypeConverter typeConverter;
+    ConversionTarget target(context);
+
+    target.addDynamicallyLegalOp<P4HIR::OverloadSetOp>([&](P4HIR::OverloadSetOp ovl) {
+        for (auto opFunc : ovl.getOps<P4HIR::FuncOp>()) return !target.isIllegal(opFunc);
+        return true;
+    });
+
+    target.addDynamicallyLegalOp<P4HIR::CaseOp>([&](P4HIR::CaseOp caseOp) {
```
asl, line 71: Looks like some copy & paste?

asl (changes requested):

include/p4mlir/Dialect/P4HIR/P4HIR_Attrs.td
asl: context is the object that owns types (and attributes). Sometimes we can deduce it from somewhere else (e.g. from compound types we certainly need to use the same context as the inner type). So, the builder w/o context is just a sugar, in general we should always have a build with context available.

lib/Transforms/RemoveAliases.cpp
xl4624: Sorry, missed the fact that aliases can't be used in switch/case operations

lib/Transforms/RemoveAliases.cpp
xl4624: Added `remove-aliases-error.mlir`, and fixed the typeswitch

PR #217, opened 2025-07-29
Implement select cases simplifications (#204)

### LINKED ISSUE #204: Add select cases simplification
Implement SimplifySelectList / SimplifySelectCases / ReplaceSelectRange from P4C

mtsamis:
This PR implements the various transition_select optimizations from ticket #204.

Four of the five transformations (`SimplifySelectList` / `ReplaceSelectRange` / `RemoveSelectBooleans` / `SingleArgumentSelect`) are implemented as patterns of a new `SimplifySelect` optimization pass, while the equivalent of `SimplifySelectCases` is implemented as `ParserTransitionSelectOp` canonicalizations.

To aid implementing these optimizations and also simplify the IR, there is an initial commit that  makes `p4hir.transition_select` and the corresponding yield statements have variadic arguments. This was done because having either a single argument or a tuple of arguments and set products made all passes more complicated and required code duplication to handle one vs many arguments.

There is also an additional commit that refactors `CastOp` constant folding and fixes various signedness bugs around it.

### DIFF
--- a/lib/Transforms/SimplifySelect.cpp
+++ b/lib/Transforms/SimplifySelect.cpp
@@ -1,3 +1,8 @@
+#include <optional>
+#include <string>
+#include <tuple>
+#include <vector>
+
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/ErrorHandling.h"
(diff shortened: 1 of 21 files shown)


lib/Transforms/SimplifySelect.cpp
```diff
@@ -0,0 +1,700 @@
...
+                        return false;
+                }
+            }
+        }
+
+        if (shape.size() != keys.size() && !isKeysUniversalSet)
+            return false;
+
+        bool isShapePrimitive = (shape.size() == 1)
+                                && !shape.front().getDefiningOp<P4HIR::TupleOp>();
```
asl, line 123: is this really how `clang-format` did? If not, will you please reformat the things? :)

asl (commented):
See comments

lib/Transforms/SimplifySelect.cpp
mtsamis: No, because I didn't use `clang-format` :sweat_smile:
I now properly formatted all the code in the last push.

lib/Transforms/SimplifySelect.cpp
```diff
@@ -0,0 +1,686 @@
...
+    if (replaceRanges)
+        patterns.add<ReplaceRanges>(patterns.getContext(), replaceRangesCaseCountLimit);
+    if (concatArgs) patterns.add<MakeSingleArgument>(patterns.getContext());
+
+    // Collect operations and apply patterns.
+    llvm::SmallVector<mlir::Operation *, 16> ops;
+    getOperation()->walk([&](P4HIR::ParserTransitionSelectOp op) { ops.push_back(op); });
+
+    mlir::GreedyRewriteConfig config;
+    config.strictMode = mlir::GreedyRewriteStrictness::ExistingOps;
```
asl, line 677: What if we'd have bool inside tuple as a key? Won't this prevent application of second pattern to the newly created op? Should we use `ExistingAndNewOps` here?

asl (commented):
See question above

mtsamis:
Latest push fixes and improves the code for `FlattenTuples`:

Previously both `shape` and `keys` arguments in `flattenValues` were `mlir::ValueRange` and we unpacked `shape` by looking for `TupleOp`. This was incorrect, since a tuple can be something other than a `TupleOp` (e.g. a tuple valued argument). Apart from resulting in bugs it also didn't make sense for `shape` to be a `ValueRange`.

In the latest code `shape` is replaced by the appropriate `TypeRange` and we unpack the actual type. The function now creates `p4hir.tuple_extract` operations to unpack values of tuple type that are not the result of a `TupleOp`.

The test case was extended with the previously failing cases.

mtsamis:
> > In the latest code `shape` is replaced by the appropriate `TypeRange` and we unpack the actual type. The function now creates `p4hir.tuple_extract` operations to unpack values of tuple type that are not the result of a `TupleOp`.
> 
> Can you always do this? And let canonicalizer fold the extracts?

Done by also implementing TupleExtractOp::fold (same as with StructExtractOp).

mtsamis:
The last force-pushed revision of this also modified the SetAttr::getConstantValue.
Instead of a templated AttrT type it handles all Attr types and returns the appropriate constant value:

```c++
llvm::APInt getConstantValue() {
  assert(getKind() == SetKind::Constant && "Expected Constant kind set.");
  assert(getMembers().size() == 1 && "Expected single member set.");
  auto attr = getMembers()[0];
  if (auto intAttr = mlir::dyn_cast<P4HIR::IntAttr>(attr)) {
    return intAttr.getValue();
  } else if (auto boolAttr = mlir::dyn_cast<P4HIR::BoolAttr>(attr)) {
    return llvm::APInt(1, static_cast<uint64_t>(boolAttr.getValue()));
  } else {
    auto enumFieldAttr = mlir::cast<P4HIR::EnumFieldAttr>(attr);
    auto enumType = mlir::cast<P4HIR::SerEnumType>(enumFieldAttr.getType());
    auto field = enumFieldAttr.getField().getValue();
    return enumType.valueOf<P4HIR::IntAttr>(field).getValue();
  }
}
```

This fixed a crash when we had a SerEnum type in a select during the `MakeSingleArgument` pattern.

mtsamis:
This is both outdated and implemented in a bad way. I will open PRs with new implementation for the same set of features.

Issue #99, opened 2025-03-06
Remove conditional control flow from parsers


asl:
See `RemoveParserControlFlow` p4c frontend pass

elhewaty:
Hi @asl, Sorry for long absence, I had to deal with a lot of things. I can find `RemoveParserControlFlow` in p4c repo, can you provide a link to it?

asl:
Well, it is pretty strange that you were unable to find it in the p4c repo. Have you checked in https://github.com/p4lang/p4c/blob/main/frontends/p4/parserControlFlow.h and https://github.com/p4lang/p4c/blob/main/frontends/p4/parserControlFlow.cpp ?

PR #220, opened 2025-08-06
Implement new remove-parser-control-flow optimization pass (#99)

mtsamis:
This PR implements the equivalent of P4C's `RemoveParserControlFlow` pass. It replaces if statements inside parser states with newly introduced states and transitions.

Some infrastructure for this pass is generally useful and will be helpful in implementing subparser/control inlining. As such, it has been placed in a separate IRUtils class.

### DIFF
--- a/include/p4mlir/Transforms/Passes.h
+++ b/include/p4mlir/Transforms/Passes.h
@@ -26,6 +26,7 @@ namespace P4::P4MLIR {
 #define GEN_PASS_DECL_SIMPLIFYPARSERS
 #define GEN_PASS_DECL_SERENUMELIMINATION
 #define GEN_PASS_DECL_REMOVEALIASES
+#define GEN_PASS_DECL_REMOVEPARSERCONTROLFLOW
 #include "p4mlir/Transforms/Passes.h.inc"
 
 std::unique_ptr<mlir::Pass> createPrintParsersGraphPass();
@@ -33,6 +34,7 @@ std::unique_ptr<mlir::Pass> createSimplifyParsersPass();
 std::unique_ptr<mlir::Pass> createFlattenCFGPass();
 std::unique_ptr<mlir::Pass> createSerEnumEliminationPass();
 std::unique_ptr<mlir::Pass> createRemoveAliasesPass();
+std::unique_ptr<mlir::Pass> createRemoveParserControlFlowPass();
 
 #define GEN_PASS_REGISTRATION
 #include "p4mlir/Transforms/Passes.h.inc"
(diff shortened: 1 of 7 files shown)


include/p4mlir/Transforms/IRUtils.h
```diff
@@ -0,0 +1,144 @@
...
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/ErrorHandling.h"
+#include "p4mlir/Dialect/P4HIR/P4HIR_Ops.h"
+#include "p4mlir/Transforms/Passes.h"
+
+namespace P4::P4MLIR::detail {
+
+class IRUtils {
```
asl, line 13: Do you really need these to be in class rather than in namespace?

test/Transforms/Passes/remove-parser-ifs.mlir
```diff
@@ -0,0 +1,252 @@
...
+      p4hir.parser_accept
+    }
+    p4hir.state @reject {
+      p4hir.parser_reject
+    }
+    p4hir.transition to @p4::@start
+  }
+
+  // CHECK-LABEL: p4hir.parser @p5
+  p4hir.parser @p5(%arg0: !b8i {p4hir.dir = #p4hir<dir in>, p4hir.param_name = "packet"}, %arg1: !p4hir.ref<!b8i> {p4hir.dir = #p4hir<dir out>, p4hir.param_name = "ipv4"})() {
```
asl, line 207: Can we, please, have test with switches and loops?

mtsamis: As far as I checked with P4C, switch statements (even with integer key) are not allowed inside parsers.
Loops are not allowed and also cannot be transformed.

So I believe there's nothing else to do for parsers.

asl: > As far as I checked with P4C, switch statements (even with integer key) are not allowed inside parsers.

This is just the small technical limitation. This is expected to be further extension of P4 spec. Nothing prevents us to have switches in P4HIR

mtsamis: Ok, it didn't occur to me. I'll follow up with the implementation then.

mtsamis:
I have completely remade the implementation. Some key changes:
 - Introduce `SplitStateRewriter` utility class that aids in transforming parser states.
 - Introduce `TransitionSelectBuilder` utility class to easily build transition select statements.
 - Complete rewrite of the pass using OpRewritePatterns for both if and switch statements
 - Added new test with nested switch and if statements.

PR #164, opened 2025-04-14
Add BinaryOp verifier and folder

xl4624:

PR #227, opened 2025-08-21
Add folding and canonicalization patterns for BinOp, CmpOp, SliceOp

mtsamis:
This PR adds folding and canonicalization for BinOp, CmpOp and SliceOp. It also introduces useful helpers and simplifies existing code.

Some of the patterns were inspired by the `arith` base dialect. It also includes the BinOp verifier from #164.

There are potentially more patterns to be added, but this should be a good start.

### DIFF
--- a/lib/Dialect/P4HIR/P4HIR_Ops.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Ops.cpp
@@ -61,6 +61,7 @@ static P4HIR::IntAttr subIntegerAttrs(mlir::PatternRewriter &builder, mlir::Valu
     return applyToIntegerAttrs(builder, res, lhs, rhs, std::minus<APInt>());
 }
 
+// Helper function to check if `attr` is an integer constant that equals the signed amount `val`.
 static bool isIntegerAttr(mlir::Attribute attr, int64_t val) {
     if (auto intAttr = mlir::dyn_cast_if_present<P4HIR::IntAttr>(attr)) {
         llvm::APInt intVal = intAttr.getValue();
@@ -70,6 +71,7 @@ static bool isIntegerAttr(mlir::Attribute attr, int64_t val) {
     return false;
 }
 
+// Helper function to create a integer `attr` that represents `val` sign-extended to match `type`.
 static mlir::Attribute getIntegerAttr(mlir::Type type, int64_t val) {
     if (auto intType = mlir::dyn_cast<P4HIR::BitsType>(type))
         return P4HIR::IntAttr::get(type, llvm::APInt(intType.getWidth(), val, true));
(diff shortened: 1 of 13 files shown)


asl:
> While developing this, I was wondering whether it would make sense to make shift left/right a binop kind?

Well, this breaks important invariant that types of rhs and lhs are same and corresponds to the result one, plus there are additional constraints on the types of arguments. So I decided to make these separate

mtsamis:
> > While developing this, I was wondering whether it would make sense to make shift left/right a binop kind?
> 
> Well, this breaks important invariant that types of rhs and lhs are same and corresponds to the result one, plus there are additional constraints on the types of arguments. So I decided to make these separate

Fair enough, and that's why I mentioned it being justified. My thought was that if we implement the constraints in e.g. an op verifier then we could duplicate the other parts, assuming that the verified will do an equivalent job.

In any case, it's not really important, I just wanted to know whether other reasons existed too.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -300,6 +345,261 @@ void P4HIR::BinOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+        return emitOpError("Unknown BinOp kind");
+    }
+
+    return emitOpError("Unknown BinOp result type");
+}
+
+OpFoldResult P4HIR::BinOp::fold(FoldAdaptor adaptor) {
+    P4HIR::BinOpKind kind = getKind();
+
+    bool isCommutative = (kind == P4HIR::BinOpKind::Add) || (kind == P4HIR::BinOpKind::AddSat) ||
```
asl, line 441: could be moved to binop method

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -300,6 +345,261 @@ void P4HIR::BinOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+    } else if (kind == P4HIR::BinOpKind::Mod) {
+        // mod(a, 1) -> 0
+        if (isIntegerAttr(adaptor.getRhs(), 1)) return getIntegerAttr(getType(), 0);
+
+        // mod(0, a) -> 0
+        if (isIntegerAttr(adaptor.getLhs(), 0)) return getIntegerAttr(getType(), 0);
+
+        return foldIntBinop(std::modulus<llvm::APSInt>{});
+    } else if (kind == P4HIR::BinOpKind::And || kind == P4HIR::BinOpKind::Or) {
+        int64_t neutralVal = (kind == P4HIR::BinOpKind::And) ? int64_t(-1) : int64_t(0);
```
asl, line 528: Are these correct? Should it be apint?

mtsamis:
Reviews addressed in the last commit, along with some more comments and improvements for `constFoldBinOp`.

asl (commented):
See few small comments and suggestions

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -296,10 +343,287 @@ OpFoldResult P4HIR::UnaryOp::fold(FoldAdaptor adaptor) {
 // BinaryOp
 //===----------------------------------------------------------------------===//
 
+static bool isCommutative(P4HIR::BinOpKind kind) {
```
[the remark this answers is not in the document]
mtsamis: Indeed, my bad. Adjusted in last commit.

PR #240, opened 2025-10-24
Add enum elimination pass

mtsamis:
Implements #201, roughly based on the existing PR #209 with some differences:

 - Implement `EnumRepresentationModel` as an external model rather than attaching the interface to `EnumType`. Downstream targets that may want different behaviour are expected to call `P4HIR::EnumType::attachInterface<SomeOtherEnumRepresentationModel>(*getContext());` after loading the P4HIR Dialect.
 - Moved `EnumFieldAttr` conversion to ConversionPatterns as it is generally useful functionality.
 - Simplified pass implementation and removed the explicit `CaseOp` legality callback by generalizing `markUnknownOpDynamicallyLegal` for attributes.

### DIFF
--- a/lib/Conversion/P4HIRToCoreLib.cpp
+++ b/lib/Conversion/P4HIRToCoreLib.cpp
@@ -131,24 +131,6 @@ struct CallMethodOpConversionPattern : public OpConversionPattern<P4HIR::CallMet
     }
 };
 
-}  // end anonymous namespace
-
-class P4CoreLibTypeConverter : public P4HIRTypeConverter {
- public:
-    P4CoreLibTypeConverter() {
-        addConversion([&](P4HIR::ExternType extType) -> std::optional<Type> {
-            if (!extType.hasAnnotation("corelib")) return std::nullopt;
-
-            if (extType.getName() == "packet_in")
-                return P4CoreLib::PacketInType::get(extType.getContext());
-            else if (extType.getName() == "packet_out")
-                return P4CoreLib::PacketOutType::get(extType.getContext());
-
-            return std::nullopt;
-        });
-    }
-};
-
 void LowerToP4CoreLib::runOnOperation() {
     MLIRContext &context = getContext();
     mlir::ModuleOp module = getOperation();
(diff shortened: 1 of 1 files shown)


mtsamis:
Some further thoughts:

 - The `markUnknownOpDynamicallyLegal` provided here is generally useful; we already have some copy-pasted version of it in other passes. I would suggest to provide this as a utility static function in ConversionPatterns.
 - This has become very similar to the `SerEnumElimination` pass, the only difference being the type and attribute conversion part. I considered consolidating both passes under a single `EnumElimination` pass but was unsure whether that would be good design. If `markUnknownOpDynamicallyLegal` is factored out at least we can simplify `SerEnumElimination` in a similar manner.
 - Can someone provide context about the need of the `OverloadSetOp` legality pattern need? Shouldn't the nested Functions be handled already by `FunctionOpInterface` clause in the `markUnknownOpDynamicallyLegal` callback? If so let's remove it, including all the other places.

These could be made a separate PR based on feedback.

include/p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.td
```diff
@@ -176,4 +175,35 @@ def HasDefaultValue : TypeInterface<"HasDefaultValue"> {
...
+      /*args=*/        (ins)
+    >,
+    InterfaceMethod<
+      /*desc=*/        "Whether this enum should be converted to serializable form",
+      /*returnType=*/  "bool",
+      /*methodName=*/  "shouldConvert",
+      /*args=*/        (ins)
+    >,
+    InterfaceMethod<
+      /*desc=*/        "Get the numeric encoding for a specific enum field",
```
asl: I am wondering, do we have any restrictions on `getUnderlyingType`? Can it return, say `p4hir.string`? Given the current interface – yes. Then `getEncoding` likely should be changed? Looks like the intention is that `getUnderlyingType` would always return `BitsType`, no? The it should be explicitly specified this way

asl:
> The markUnknownOpDynamicallyLegal provided here is generally useful; we already have some copy-pasted version of it in other passes. I would suggest to provide this as a utility static function in ConversionPatterns.

Yes!

Issue #100, opened 2025-03-06
Implement parser inlining


asl:
Decide if we'd go via common MLIR inlining or we'd need to implement something custom from scratch. Technically we'd just split the state, inline parser body and locals and stitch the `accept` in subparser with the next state

xl4624:
## RFC

The goal here is to inline any subparser calls, and we can implement this by following these steps:
1. Look for `ApplyOp` calls, check that their parent is `ParserStateOp` and that their `callee` attribute resolves to a `ParserOp`
2. Split the caller's parser state into two by creating the post-subparser apply parser state
3. Move any operations from after the `ApplyOp` into this new state (including the transition terminator)
4. Now the original state's terminator is a transition to the start state of the inlined parser
5. Then we can use `IRMapping` to map the subparser's parameters with the `ApplyOp`'s operands (same for ctor)
6. We can also also make a copy of any subparser locals by appending them to the caller parser's entry block and RAUW
7. To handle the subparser's states, we can make a clone of each state and move them inside the caller parser
8. We can keep any `parser_reject'`s, but for `parser_accept` we replace with the split state from step 2

Here is an example pass:

Before
```mlir
p4hir.parser @subparser(%arg0: !p4hir.bit<8> {...})() {
  %c0 = p4hir.const #int0_b8i
  %local_var = p4hir.variable ["local_flag"] : !p4hir.bool
  p4hir.state @start {
    %cond = p4hir.cmp(eq, %arg0, %c0) : !p4hir.bit<8>, !p4hir.bool
    p4hir.assign %cond, %local_var : <!p4hir.bool>
    %read_cond = p4hir.read %local_var : <!p4hir.bool>
    p4hir.transition_select %read_cond : !p4hir.bool {
      p4hir.select_case { ... } to @subparser::@accept
      p4hir.select_case { ... } to @subparser::@reject
    }
  }
  p4hir.state @accept { p4hir.parser_accept }
  p4hir.state @reject { p4hir.parser_reject }
  p4hir.transition to @subparser::@start
}

p4hir.parser @parser(%arg1: !p4hir.bit<8> {...})() {
  p4hir.instantiate @subparser as @sp
  // ... other states
  p4hir.state @process {
    %c42 = p4hir.const #int42_b8i
    p4hir.apply @sp(%c42) : (!p4hir.bit<8>) -> ()
    %c10 = p4hir.const #p4int10_b8i
    // ...
    p4hir.parser_accept
  }
[...]

PR #241, opened 2025-10-24
Implement parser inlining

mtsamis:
This PR implements parser inlining (#100).

The changes include generic infrastructure found in IRUtils and an inlining pass based on MLIR `InliningUtils` and the new `SplitStateRewriter` to handle parser rewrites. The implementation makes sure to leave the IR in its original state if the inlining fails at any part (e.g. due to an apply call being inside an if statement).

Although not necessary for correctness, parser locals are shared per instantiation to avoid creating too many new locals that will be hard to eliminate later on.

A number of tests are added, including the parser inlining tests from P4C. The latter are there only to check against potential crashes.

### DIFF
--- a/lib/Transforms/InlineParsers.cpp
+++ b/lib/Transforms/InlineParsers.cpp
@@ -1,0 +1,273 @@
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/ErrorHandling.h"
+#include "llvm/Support/FormatVariadic.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Inliner.h"
+#include "mlir/Transforms/InliningUtils.h"
+#include "p4mlir/Dialect/P4HIR/P4HIR_Ops.h"
+#include "p4mlir/Transforms/IRUtils.h"
+#include "p4mlir/Transforms/Passes.h"
+
+#define DEBUG_TYPE "p4hir-inline-parsers"
+
+namespace P4::P4MLIR {
+#define GEN_PASS_DEF_INLINEPARSERS
+#include "p4mlir/Transforms/Passes.cpp.inc"
+}  // namespace P4::P4MLIR
+
+using namespace P4::P4MLIR;
+
+namespace {
+struct InlineParsersPass : public P4::P4MLIR::impl::InlineParsersBase<InlineParsersPass> {
+    InlineParsersPass() = default;
+    void runOnOperation() override;
+
+ private:
+    // Given a subparser instantiation `instOp` find all its apply calls and try to inline them.
+    // Leaves the IR unchanged in case of failure.
+    mlir::LogicalResult inlineSubparser(P4HIR::InstantiateOp instOp, mlir::RewriterBase &rewriter) {
+        P4HIR::ParserOp calleeParser = instOp.getCalleeParser();
+        if (!calleeParser) return mlir::failure();
+
+        auto parser = instOp->getParentOfType<P4HIR::ParserOp>();
+        auto calleeRegion = calleeParser.getCallableRegion();
+        auto instName = instOp.getSymName();
+
+        if (!instOp.getArgOperands().empty())
+            return rewriter.notifyMatchFailure(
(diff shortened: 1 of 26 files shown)


fruffy (approved):
The lack of E2E testing and reference files makes it quite hard to tell whether the resulting semantic behavior is correct. Would be really nice to have a packet testing framework for this at some point.

lib/Transforms/IRUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return escapingBlocks;
+}
+
+// Promote `val` to a local variable.
+// The rewriter's insertion point is the location where the new variable is created.
+void promoteValToVar(mlir::RewriterBase &rewriter, mlir::Value val, const BlockSet &escapingBlocks,
+                     llvm::StringRef varName = "promoted_local") {
+    // Create new variable to hold `val`.
+    auto newVar = rewriter.create<P4HIR::VariableOp>(
+        val.getLoc(), P4HIR::ReferenceType::get(val.getType()), varName);
```
asl: Add `init` attribute as the variable will be immediately initialized.

lib/Transforms/IRUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+                if (!escapingBlocks.empty()) promoteValToVar(rewriter, val, escapingBlocks);
+            }
+        }
+    }
+}
+
+P4HIR::ParserStateOp IRUtils::createSubState(mlir::RewriterBase &rewriter,
+                                             P4HIR::ParserStateOp state, llvm::StringRef suffix) {
+    auto parser = state->getParentOfType<P4HIR::ParserOp>();
+    std::string basename = (llvm::Twine(state.getSymName()) + "_" + suffix).str();
```
asl: Do you really need `Twine` here? I'd suggest using `SmallString` here with pre-allocated buffer

lib/Transforms/IRUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::string name = basename;
+
+    size_t counter = 0;
+    while (parser.lookupSymbol(name)) name = basename + "_" + std::to_string(counter++);
```
asl: Use `mlir::SymbolTable::generateSymbolName`. There are examples of usage in `P4C/translate.cpp`

lib/Transforms/InlineParsers.cpp
```diff
@@ -0,0 +1,252 @@
...
+                inliningRewriters.emplace_back(
+                    std::make_unique<IRUtils::SplitStateRewriter>(rewriter, applyOp));
+        });
+
+        // We are calling the inliner once per apply, but some operations must/should be cloned
+        // once per instantiation. The IRMapper argument in InliningUtils cannot be used to
+        // map whole operations within the cloned region so we must take care to de-duplicate parser
+        // locals ourselves.
+        // For each operation in the callee's top level block, `declMapping` holds an operation if
+        // we must remmap the I-th operation and nullptr otherwise. We have to use a vector of
```
asl:
```suggestion
        // we must remap the I-th operation and nullptr otherwise. We have to use a vector of
```

lib/Transforms/InlineParsers.cpp
```diff
@@ -0,0 +1,252 @@
...
+        }
+    };
+};
+
+}  // namespace
+
+void InlineParsersPass::runOnOperation() {
+    mlir::IRRewriter rewriter(&getContext());
+
+    // P4 guarantees that subparser definitions come before their use so inlining them
```
asl: This is true for P4 input source. But technically since we're referring things via symbols and symbol tables this might be not true anymore and this is not enforced in P4HIR. I would explicitly have a worklist and iterate until it is not empty

lib/Transforms/InlineParsers.cpp
```diff
@@ -0,0 +1,252 @@
...
+                for (mlir::Operation &op : llvm::make_early_inc_range(block)) {
+                    proccessInlinedOp(index, &op);
+                    index++;
+                }
+            }
+        }
+
+        // Move inlined operation `op` to its desired position.
+        void proccessInlinedOp(size_t index, mlir::Operation *op) {
+            // Remmap parser locals if needed.
```
asl:
```suggestion
            // Remap parser locals if needed.
```

mtsamis:
Superseded by https://github.com/p4lang/p4mlir-incubator/pull/246.

PR #245, opened 2025-10-31
Replace P4::RemoveReturns with MLIR-based implementation

mtsamis:
This PR replaces `P4::RemoveReturns` in translator with new MLIR-based operations and a corresponding "lowering" pass. At the same time, this brings support for `continue` and `break` statements.

The new operations introduced are `soft_return`, `soft_break` and `soft_continue` which are not terminators (a normal `return` terminator is expected where appropriate). The `RemoveSoftCF` will eliminate these operations and replace them with appropriate guards or by moving operations to suitable places. 

Compared to `P4::RemoveReturns`, the MLIR-based implementation aims to be simpler while also generating better IR.

### DIFF
--- a/test/Transforms/Passes/remove-soft-cf-composite.mlir
+++ b/test/Transforms/Passes/remove-soft-cf-composite.mlir
@@ -1,4 +1,4 @@
-// RUN: p4mlir-opt --p4hir-remove-soft-cf --canonicalize %s | FileCheck %s
+// RUN: p4mlir-opt --pass-pipeline='builtin.module(p4hir.func(p4hir-remove-soft-cf, canonicalize))' %s | FileCheck %s
 
 // Test combination of all soft control flow statements.
 
--- a/test/Transforms/Passes/remove-soft-cf-control.mlir
+++ b/test/Transforms/Passes/remove-soft-cf-control.mlir
@@ -1,4 +1,4 @@
-// RUN: p4mlir-opt --p4hir-remove-soft-cf %s | FileCheck %s
+// RUN: p4mlir-opt --pass-pipeline='builtin.module(p4hir.control(p4hir-remove-soft-cf))' %s | FileCheck %s
 
 !b16i = !p4hir.bit<16>
 !S = !p4hir.struct<"S", f: !b16i>
--- a/test/Transforms/Passes/remove-soft-cf-if-switch.mlir
+++ b/test/Transforms/Passes/remove-soft-cf-if-switch.mlir
@@ -1,4 +1,4 @@
-// RUN: p4mlir-opt --p4hir-remove-soft-cf --canonicalize %s | FileCheck %s
+// RUN: p4mlir-opt --pass-pipeline='builtin.module(p4hir.func(p4hir-remove-soft-cf, canonicalize), p4hir.overload_set(p4hir.func(p4hir-remove-soft-cf, canonicalize)))' %s | FileCheck %s
 
 !b8i = !p4hir.bit<8>
 #int-56_b8i = #p4hir.int<200> : !b8i
--- a/test/Transforms/Passes/remove-soft-cf-loop.mlir
+++ b/test/Transforms/Passes/remove-soft-cf-loop.mlir
@@ -1,4 +1,4 @@
-// RUN: p4mlir-opt --p4hir-remove-soft-cf --canonicalize %s | FileCheck %s
+// RUN: p4mlir-opt --pass-pipeline='builtin.module(p4hir.func(p4hir-remove-soft-cf, canonicalize))' %s | FileCheck %s
 
 !b8i = !p4hir.bit<8>
 #int0_b8i = #p4hir.int<0> : !b8i


include/p4mlir/Transforms/Passes.td
```diff
@@ -157,4 +157,44 @@ def EnumElimination : Pass<"p4hir-enum-elimination", "mlir::ModuleOp"> {
...
+
+    Will be transformed to something like this:
+    ```
+    bit<8> func(bit<8> a) {
+      bit<8> result;
+      if (a > 200) {
+        result = 1;
+      } else {
+        if (a > 100) {
+          retult = 2;
```
asl:
```suggestion
          return = 2;
```

include/p4mlir/Transforms/Passes.td
```diff
@@ -157,4 +157,44 @@ def EnumElimination : Pass<"p4hir-enum-elimination", "mlir::ModuleOp"> {
...
+def RemoveSoftCF : Pass<"p4hir-remove-soft-cf", "mlir::ModuleOp"> {
+  let summary = "Convert functions and controls to SESE form.";
+  let description = [{
+    Convert functions and controls to Single-Entry-Single-Exit (SESE) form by restructuring
+    code and eliminating return, break and continue statements.
+
+    For example code like this:
+    ```
+    bit<8> func(bit<8> a) {
+      if (a > 200) { return 1; }
```
asl: Will you please also add an example of transformation when `else` is present?

lib/Transforms/RemoveSoftCF.cpp
```diff
@@ -0,0 +1,377 @@
...
+ private:
+    P4HIR::VariableOp guardVar;
+};
+
+// Helper struct to replace all soft control operations within a function-like operation.
+// This is done by moving operations or introducing guard variables that control execution of
+// subsequent operations. A core observation used is that if we're currently adjusting an operation
+// at some "nest level" X, we can place all subsequent operations at nest level Y >= X while
+// guaranteeing the original semantics are preserved (we never need to introduce a guarded block at
+// an outter level). Operations are either moved to a more nested block or to a guard block that is
```
asl:
```suggestion
// an outer level). Operations are either moved to a more nested block or to a guard block that is
```

lib/Transforms/RemoveSoftCF.cpp
```diff
@@ -0,0 +1,377 @@
...
+        assert(region->hasOneBlock() && "Expected region with one block");
+        mlir::Block *block = &region->front();
+
+        rewriter.setInsertionPointToStart(block);
+        returnGuard.init(rewriter, loc, "return_guard");
+
+        // Create a variable to store the final return if needed.
+        if (returnType && !mlir::isa<P4HIR::VoidType>(returnType))
+            returnVar = rewriter.create<P4HIR::VariableOp>(
+                loc, P4HIR::ReferenceType::get(returnType), "return_value");
```
asl: Add `init` here. In the future we can enforce that all such variables should be immediately initialized.

mtsamis: Ok. I thought that `init` doesn't hold here, but we can probably guarantee that it will be initialized before the first use.

asl: Well... if it would be used before initialization, then this is a problem.

mtsamis: I mean yes, it shouldn't happen in normal circumstances, except for "malformed input" like problematic MLIR input. I assume that shouldn't really bother us though.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -806,11 +806,21 @@ void P4HIR::VariableOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
 }
 
 LogicalResult P4HIR::VariableOp::canonicalize(P4HIR::VariableOp op, PatternRewriter &rewriter) {
+    auto users = op->getUsers();
+    auto firstNonAssignOp =
+        llvm::find_if(users, [](auto *user) { return !mlir::isa<P4HIR::AssignOp>(user); });
+
+    if (firstNonAssignOp == users.end()) {
+        // Completely remove variable if it is only written to.
```
asl: This is just an optimization here, right? The pass itself does not depend on this?

include/p4mlir/Transforms/Passes.td
```diff
@@ -157,4 +157,51 @@ def EnumElimination : Pass<"p4hir-enum-elimination", "mlir::ModuleOp"> {
   let dependentDialects = ["P4MLIR::P4HIR::P4HIRDialect"];
 }
 
+//===----------------------------------------------------------------------===//
+// Remove Soft Control Flow
+//===----------------------------------------------------------------------===//
+
+def RemoveSoftCF : Pass<"p4hir-remove-soft-cf", "mlir::ModuleOp"> {
```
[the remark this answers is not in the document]
mtsamis: I also made `overload_set` have the `IsolatedFromAbove` trait so we can run the pass for functions nested there.

Issue #208, opened 2025-06-16
Implement control inliner


asl:

asl:
This was implemented in #246

PR #246, opened 2025-11-07
Implement parser and control inlining

mtsamis:
This PR implements parser and control inlining (https://github.com/p4lang/p4mlir-incubator/issues/100, https://github.com/p4lang/p4mlir-incubator/issues/208).

Contrast to the original parser inlining PR (https://github.com/p4lang/p4mlir-incubator/pull/241), this doesn't use mlir's `InliningUtils`. The reason is that `InliningUtils` is designed for function-like object whereas parsers and controls are not function-like at the IR level. Trying to use that caused more issues than it solved, especially for controls.

This new PR adds a helper class `InstantiateOpInlineHelper` which provides much of the shared logic for inlining and then two separate subclasses for parser and control inlining. Parser inlining uses `SplitStateRewriter` as before and the new implementation is simpler in comparison.

For control inlining the most important aspect is handling of `p4hir.parser_local`s, which may need to be promoted from values to variable references. All aspects of that can be found in the implementation comments and relevant tests.

### DIFF
--- a/lib/Dialect/P4HIR/P4HIR_Ops.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Ops.cpp
@@ -840,10 +840,12 @@ LogicalResult P4HIR::VariableOp::canonicalize(P4HIR::VariableOp op, PatternRewri
         if (!mlir::isa<ReadOp>(user) || user->isBeforeInBlock(uniqueAssignOp)) return failure();
     }
 
-    // Remove the assign op and replace all reads with the new assigned var op.
+    // Replace all reads with the assigned value and remove the assignment.
     mlir::Value assignedValue = uniqueAssignOp.getValue();
+    for (auto *user : llvm::make_early_inc_range(users))
+        if (user != uniqueAssignOp)
+            rewriter.replaceOp(user, assignedValue);
     rewriter.eraseOp(uniqueAssignOp);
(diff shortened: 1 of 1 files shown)


lib/Transforms/InlineParserControl.cpp
```diff
@@ -0,0 +1,466 @@
...
+    InstantiateOpInlineHelper(mlir::RewriterBase &rewriter, P4HIR::InstantiateOp instOp)
+        : rewriter(rewriter), instOp(instOp) {}
+
+    mlir::LogicalResult init() {
+        callee = instOp.getCallee<OpTy>();
+        if (!callee) return mlir::failure();
+
+        if (!instOp.getArgOperands().empty())
+            return rewriter.notifyMatchFailure(instOp.getLoc(),
+                                               "Cannot inline object with constructor arguments.");
```
asl: Can't we just RAUW ctor arguments during cloning?

lib/Transforms/InlineParserControl.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+        if (!callee.getCallableRegion()->hasOneBlock())
+            return rewriter.notifyMatchFailure(instOp.getLoc(),
```
asl: Should we enforce this at control / parser level?

asl (commented):
Few first notes

asl (approved):
Thanks!

PR #269, opened 2025-11-18
Support constructor argument access in control actions

mtsamis:
Currently control constructor arguments cannot be accessed inside control actions, it results in the translator crashing.
This PR makes all constructor constants available in actions, as it is expected.

Since constructor arguments are referred to using the `ctor_param` attribute and a name, I believe we don't need to export them as control locals and can use the same IR as when they're accessed outside actions.

The constants are created eagerly in all actions, so unused arguments will be cleaned up only when the canonicalizer is ran.


mtsamis:
> We need a test ;)

Done :)

asl (approved):

PR #326, opened 2026-03-08
Add lowering support for static_assert

devalgupta404:
issue #183 
Add lowering support for static_assert in the P4HIRToCoreLib conversion.
First I evaluates the assertion condition at compile time. If the condition is a constant boolean true, the call is replaced with p4hir.const #true. If the condition evaluates to false, a diagnostic (static assertion failed) is emitted during lowering.

Covers Four Test cases 
- static_assert(true)
- static_assert(true, "message")
- static_assert(false)
- static_assert(false, "message")

I am using '--split-input-file' to run each cases in an independent module otherwise repeated definations of 'static_assert' will cause duplicate symbol errors.

### DIFF
--- a/include/p4mlir/Dialect/P4CoreLib/P4CoreLib_Ops.td
+++ b/include/p4mlir/Dialect/P4CoreLib/P4CoreLib_Ops.td
@@ -102,4 +102,24 @@ def PacketExtractVariableOp : P4CoreLib_Op<"extract_header_variable"> {
     }];
 }
 
+def StaticAssertOp : P4CoreLib_Op<"static_assert"> {
+    let summary = "Compile-time assertion";
+    let description = [{
+        Represents a compile-time assertion. The condition must eventually
+        evaluate to a constant boolean. If the condition is true, this op
+        can be folded away. If false, a separate pass will emit an error.
+        
+        Optionally includes a message string for error reporting.
+    }];
+
+    let arguments = (ins BooleanType:$cond,
+                         OptionalAttr<StrAttr>:$message);
+    let results = (outs BooleanType:$result);
+    let assemblyFormat = [{
+      $cond (`message` $message^)? attr-dict `:` type($cond) `->` type($result)
+    }];
+    
+    let hasFolder = 1;
+}
+
 #endif //P4MLIR_DIALECT_P4CORELIB_P4CORELIB_OPS_TD
--- a/lib/Conversion/P4HIRToCoreLib/P4HIRToCoreLib.cpp
+++ b/lib/Conversion/P4HIRToCoreLib/P4HIRToCoreLib.cpp
@@ -85,31 +85,31 @@ struct CallOpConversionPattern : public OpConversionPattern<P4HIR::CallOp> {
                                                              operands.getArgOperands());
             return success();
         }
-        // Handle static_assert: compile-time assertion
+        // Handle static_assert: lower to P4CoreLib::StaticAssertOp
         if (callee.getLeafReference().getValue().starts_with("static_assert")) {
             auto args = operands.getArgOperands();
         
             if (args.empty())
                 return op.emitError("static_assert requires a condition");
         
-            // Ensure the condition is a compile-time constant
-            Attribute constAttr;
-            if (!matchPattern(args[0], m_Constant(&constAttr)))
-                return op.emitError("static_assert condition must be constant");
-        
-            // Ensure the constant is boolean
-            auto boolAttr = llvm::dyn_cast<P4HIR::BoolAttr>(constAttr);
-            if (!boolAttr)
-                return op.emitError("static_assert condition must be boolean");
-        
-            // If condition is false, emit compile-time error
-            if (!boolAttr.getValue())
-                return op.emitError("static assertion failed");
-        
-            // If condition is true, replace the call with a constant true
-            auto trueAttr = P4HIR::BoolAttr::get(rewriter.getContext(), true);
-            rewriter.replaceOpWithNewOp<P4HIR::ConstOp>(op, trueAttr);
-        
+            Value cond = args[0];
+            StringAttr message = nullptr;
+            if (args.size() == 2) {
+                if (auto constOp = args[1].getDefiningOp<P4HIR::ConstOp>()) {
+                    message = llvm::dyn_cast<StringAttr>(constOp.getValue());
+                }
+                 if (!message) {
+                    op.emitWarning("static_assert message is not a compile-time constant; ignoring");
+                 }
+                
+            }
+              
+            rewriter.replaceOpWithNewOp<P4CoreLib::StaticAssertOp>(
+                op,
+                op.getResult().getType(),
+                cond,
+                message
+            );
             return success();
         }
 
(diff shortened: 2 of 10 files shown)


asl (changes requested):

devalgupta404:
Hi @asl,
I’ve updated the PR to follow the suggested approach.
i Introduced a a new p4corelib.static_assert op with clear semantics, Updated lowering to avoid enforcing compile-time constants during conversion also Implemented a dedicated EvaluateStaticAssertPass and Added folding support in the op for early canonicalization of true cases as well.

devalgupta404:
@mtsamis can you review it please

lib/Conversion/P4HIRToCoreLib/P4HIRToCoreLib.cpp
```diff
@@ -84,6 +85,33 @@ struct CallOpConversionPattern : public OpConversionPattern<P4HIR::CallOp> {
                                                              operands.getArgOperands());
             return success();
         }
+        // Handle static_assert: lower to P4CoreLib::StaticAssertOp
+        if (callee.getLeafReference().getValue().starts_with("static_assert")) {
```
mtsamis, line 89: This is an insufficient check (e.g. there could be a user named function `static_assert_condition` that would pass this). static assert will be only present in an overload set that looks like this:
```
  p4hir.overload_set @static_assert {
    p4hir.func @static_assert_0(!p4hir.bool {p4hir.dir = #undir, p4hir.param_name = "check"}, !string {p4hir.dir = #undir, p4hir.param_name = "message"}) -> !p4hir.bool
    p4hir.func @static_assert_1(!p4hir.bool {p4hir.dir = #undir, p4hir.param_name = "check"}) -> !p4hir.bool
  }
```
So you should get the overload_set and then check if it's `static_assert`.

lib/Conversion/P4HIRToCoreLib/P4HIRToCoreLib.cpp
```diff
@@ -84,6 +85,33 @@ struct CallOpConversionPattern : public OpConversionPattern<P4HIR::CallOp> {
...
+            Value cond = args[0];
+            StringAttr message = nullptr;
+            if (args.size() == 2) {
+                if (auto constOp = args[1].getDefiningOp<P4HIR::ConstOp>()) {
+                    message = llvm::dyn_cast<StringAttr>(constOp.getValue());
+                }
+                 if (!message) {
+                    op.emitWarning("static_assert message is not a compile-time constant; ignoring");
+                 }
+
```
mtsamis, line 104: Remove empty line, plus formatting above is off, have you run clang-format?
Also, as the llvm code styling guideline suggests, please don't use brackets for simple one-line if statements.

lib/Conversion/P4HIRToCoreLib/P4HIRToCoreLib.cpp
```diff
@@ -84,6 +85,33 @@ struct CallOpConversionPattern : public OpConversionPattern<P4HIR::CallOp> {
...
+            if (args.empty())
+                return op.emitError("static_assert requires a condition");
+        
+            Value cond = args[0];
+            StringAttr message = nullptr;
+            if (args.size() == 2) {
+                if (auto constOp = args[1].getDefiningOp<P4HIR::ConstOp>()) {
+                    message = llvm::dyn_cast<StringAttr>(constOp.getValue());
+                }
+                 if (!message) {
```
mtsamis, line 101: Is this allowed to happen? (i.e. a non constant string argument).
If not, use assert/llvm_unreachable

lib/Dialect/P4CoreLib/P4CoreLib_Ops.cpp
```diff
@@ -14,6 +15,27 @@ void P4CoreLib::PacketLookAheadOp::getAsmResultNames(mlir::OpAsmSetValueNameFn s
     setNameFn(getResult(), "lookahead");
 }
 
+OpFoldResult P4CoreLib::StaticAssertOp::fold(FoldAdaptor adaptor) {
```
mtsamis, line 18: I'd say we shouldn't have this folding pattern. It basically implements half of the pass, and creates redundancy while not providing any clear benefit as we'd have to run the pass anyway. Was this intended to be an optimization?

lib/Transforms/EvaluateStaticAssert.cpp
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    for (auto op : assertOps) {
+      Value cond = op.getCond();
+
+      auto constOp = cond.getDefiningOp<P4HIR::ConstOp>();
+      if (!constOp)
+        continue;
+
+      auto boolAttr = llvm::dyn_cast<P4HIR::BoolAttr>(constOp.getValue());
```
mtsamis, line 39: Since cond is BooleanType, let's use llvm::cast here and remove `if (!boolAttr) continue;`

lib/Transforms/EvaluateStaticAssert.cpp
```diff
@@ -0,0 +1,69 @@
...
+    auto moduleOp = getOperation();
+    bool hasError = false;
+
+    SmallVector<P4CoreLib::StaticAssertOp> assertOps;
+    moduleOp->walk([&](P4CoreLib::StaticAssertOp op) {
+      assertOps.push_back(op);
+    });
+
+    for (auto op : assertOps) {
+      Value cond = op.getCond();
```
mtsamis, line 33: No need for single use variable here, just do `op.getCond().getDefiningOp...` below

test/Conversion/P4CoreLib/static_assert.mlir
```diff
@@ -0,0 +1,44 @@
...
+// Test lowering of static_assert calls to p4corelib.static_assert op
+
+!string = !p4hir.string
+#true = #p4hir.bool<true> : !p4hir.bool
+#false = #p4hir.bool<false> : !p4hir.bool
+#undir = #p4hir<dir undir>
+
+module {
+  p4hir.func @static_assert_0(!p4hir.bool {p4hir.dir = #undir, p4hir.param_name = "check"}, !string {p4hir.dir = #undir, p4hir.param_name = "message"}) -> !p4hir.bool annotations {corelib}
+  p4hir.func @static_assert_1(!p4hir.bool {p4hir.dir = #undir, p4hir.param_name = "check"}) -> !p4hir.bool annotations {corelib}
```
mtsamis, line 12: Following the advice at the top, these must be in a overload_set, as it would be generated by translating the `core.p4` definitions.

include/p4mlir/Dialect/P4CoreLib/P4CoreLib_Ops.td
```diff
@@ -102,4 +102,24 @@ def PacketExtractVariableOp : P4CoreLib_Op<"extract_header_variable"> {
...
 }
 
+def StaticAssertOp : P4CoreLib_Op<"static_assert"> {
+    let summary = "Compile-time assertion";
+    let description = [{
+        Represents a compile-time assertion. The condition must eventually
+        evaluate to a constant boolean. If the condition is true, this op
+        can be folded away. If false, a separate pass will emit an error.
+        
+        Optionally includes a message string for error reporting.
```
mtsamis, line 112: Once the folding is removed (see other comment), let's use something along the lines of:
```
        Represents a compile-time assertion with an optional error message.
        The condition must eventually evaluate to a constant boolean.
        An error will be emitted if the condition evaluates to false.
```

mtsamis (changes requested):

PR #331, opened 2026-03-25
Add verifier checks for UnaryOp operand types

Lavanya-Malakalapalli:
This change tightens `p4hir.unary` verification by enforcing operand type constraints based on the unary operation kind.

Changes:
- require `not` to operate on `!p4hir.bool`
- require `minus`, `plus`, and `cmpl` to operate on integer-like P4HIR bit types

Also adds a negative test file covering invalid unary operations.

Validation:
- `ninja check-p4mlir` passes

Fixes: FIXME in UnaryOp verifier

### DIFF
--- a/lib/Dialect/P4HIR/P4HIR_Ops.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Ops.cpp
@@ -290,12 +290,22 @@ static P4HIR::BinOp getDefiningBinop(P4HIR::BinOpKind kind, mlir::Value val) {
 }
 
 LogicalResult P4HIR::UnaryOp::verify() {
+    auto type = getInput().getType();
+
     switch (getKind()) {
         case P4HIR::UnaryOpKind::Neg:
         case P4HIR::UnaryOpKind::UPlus:
+            if (!mlir::isa<P4HIR::BitsType,P4HIR::InfIntType>(type))
+                return emitOpError("arithmetic unary operations require integer-like type");
+            return success();
         case P4HIR::UnaryOpKind::Cmpl:
+            if (!mlir::isa<P4HIR::BitsType,P4HIR::InfIntType>(type))
+                return emitOpError("bitwise complement operations require integer-like type");
+            return success();
+
         case P4HIR::UnaryOpKind::LNot:
-            // Nothing to verify.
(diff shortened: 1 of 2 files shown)


lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -290,12 +290,19 @@ static P4HIR::BinOp getDefiningBinop(P4HIR::BinOpKind kind, mlir::Value val) {
 }
 
 LogicalResult P4HIR::UnaryOp::verify() {
+    auto type = getInput().getType();
+
     switch (getKind()) {
         case P4HIR::UnaryOpKind::Neg:
         case P4HIR::UnaryOpKind::UPlus:
         case P4HIR::UnaryOpKind::Cmpl:
+            if (!mlir::isa<P4HIR::BitsType>(type))
```
mtsamis, line 302: Neg/UPlus should probably allow infint, right?

Lavanya-Malakalapalli: Thank you!
I’ve updated the verifier to allow `InfIntType`  for` Cmpl`, `Neg `and `UPlus `,so that arithmetic unary operations and bitwise complement are both handled consistently with integer-like types. I also kept the cases separate to make the distinction between arithmetic and bitwise operations clearer.

mtsamis: InfInt is not valid to have for complement, as the result cannot be defined, see also https://p4.org/wp-content/uploads/sites/53/2024/10/P4-16-spec-v1.2.5.html#sec-varint-ops

Lavanya-Malakalapalli: I’ve updated the verifier to restrict `Cmpl `to `BitsType` only, and allow` InfIntType` only for `Neg `and `UPlus`, in line with the spec.

Just to confirm my understanding: `BitsType `in P4HIR covers fixed-width integers (both `bit<W>` and `int<W>`), while `InfIntType` represents `arbitrary-precision` integers. Does that align with the intended type modeling here?

mtsamis (approved):

PR #333, opened 2026-03-28
[P4HIR] Fix unsound CastOp chain folding (#317)

AbdallahRashed:
Cast chains A→B→C were unconditionally folded to A→C, but this changes semantics when B alters signedness before a widening cast (e.g. bit<8>→int<8>→int<16> sign-extends, but bit<8>→int<16> zero-extends).

Only fold when the second cast doesn't widen, or the first cast preserves signedness without truncating.

Fixes #317

### DIFF
--- a/lib/Dialect/P4HIR/P4HIR_Ops.cpp
+++ b/lib/Dialect/P4HIR/P4HIR_Ops.cpp
@@ -250,12 +250,57 @@ OpFoldResult P4HIR::CastOp::fold(FoldAdaptor) {
     return {};
 }
 
+/// Returns true if folding a cast chain A -> B -> C into A -> C preserves
+/// the cast semantics
+///
+/// The key unsafe case is when the second cast widens (w_B < w_C), because
+/// the extension type (zero vs sign) depends on the source's signedness.
+/// If the intermediate type B differs from A in a way that changes how
+/// the widening is performed, the fold would alter semantics.
+///
+/// Fold is safe when:
+/// - w_B >= w_C: second cast doesn't widen (truncation and reinterpretation
+///   are sign-independent), OR
+/// - w_A <= w_B AND s_A == s_B: first cast doesn't truncate and preserves
+///   signedness, so the widening in the second cast uses the same extension
+///   type as the direct A -> C cast would.
+static bool isSafeCastComposition(mlir::Type srcType, mlir::Type midType, mlir::Type dstType) {
+    if (srcType == midType || midType == dstType) return true;
+
+    auto srcBits = mlir::dyn_cast<P4HIR::BitsType>(srcType);
+    auto midBits = mlir::dyn_cast<P4HIR::BitsType>(midType);
+    auto dstBits = mlir::dyn_cast<P4HIR::BitsType>(dstType);
+
+    if (srcBits && midBits && dstBits) {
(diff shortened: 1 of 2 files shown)


AbdallahRashed:
I am opening this PR, to express my interest in GSOC: [Project 3.1: Alkali-P4MLIR: Bridging P4 and SmartNICs Through MLIR Dialect Conversion Between Alkali IR and P4MLIR]
Thank you

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -250,12 +250,57 @@ OpFoldResult P4HIR::CastOp::fold(FoldAdaptor) {
...
+/// the widening is performed, the fold would alter semantics.
+///
+/// Fold is safe when:
+/// - w_B >= w_C: second cast doesn't widen (truncation and reinterpretation
+///   are sign-independent), OR
+/// - w_A <= w_B AND s_A == s_B: first cast doesn't truncate and preserves
+///   signedness, so the widening in the second cast uses the same extension
+///   type as the direct A -> C cast would.
+static bool isSafeCastComposition(mlir::Type srcType, mlir::Type midType, mlir::Type dstType) {
+    if (srcType == midType || midType == dstType) return true;
```
mtsamis, line 268: Shouldn't be needed due to what we have in CastOp::fold

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -250,12 +250,57 @@ OpFoldResult P4HIR::CastOp::fold(FoldAdaptor) {
...
+        // Safe if the second cast doesn't widen.
+        if (wB >= wC) return true;
+
+        // Second cast widens (wB < wC). Safe only if the first cast doesn't
+        // truncate and preserves signedness, so the composed extension matches
+        // the direct A -> C extension.
+        return wA <= wB && srcBits.isSigned() == midBits.isSigned();
+    }
+
+    // For non-BitsType chains, be conservative and don't fold.
```
mtsamis, line 288: Can you please make this more complete by also handling booleans? They should be fairly easy to add and then you can also remove this comment.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -250,12 +250,57 @@ OpFoldResult P4HIR::CastOp::fold(FoldAdaptor) {
     return {};
 }
 
+/// Returns true if folding a cast chain A -> B -> C into A -> C preserves
+/// the cast semantics
+///
+/// The key unsafe case is when the second cast widens (w_B < w_C), because
+/// the extension type (zero vs sign) depends on the source's signedness.
+/// If the intermediate type B differs from A in a way that changes how
+/// the widening is performed, the fold would alter semantics.
```
mtsamis, line 259: I think this middle section of comments is overly verbose.
The safe cases are explained below and naming this the "key unsafe case" is somewhat misleading (I would say truncation is also key).

mtsamis (changes requested):
Please see some adjustments to make

mtsamis:
Opened https://github.com/p4lang/p4mlir-incubator/pull/339 based on this implementation and adding `bool`/`bit<1>` support.

PR #334, opened 2026-03-30
[Exporter] Fix parameter direction and type emission in exportArgument

p2peek:
This PR fixes parameter emission in the P4HIR → P4 exporter.

Currently, exportArgument(...) only emits annotations and misses:
- parameter direction (in/out/inout)
- parameter type
- parameter name

This change ensures that parameters are printed as valid P4 syntax, e.g.:

    in bit<32> arg0

This is a minimal fix toward completing the exporter described in PR #168.

Next steps include:
- extracting actual parameter direction from attributes
- handling parameter names properly
- improving type coverage

### DIFF
--- a/lib/Utilities/export_to_p4.cpp
+++ b/lib/Utilities/export_to_p4.cpp
@@ -4436,26 +4436,67 @@ class P4HirToP4Exporter {
     /// ```p4
     /// in bit<32> data
     /// ```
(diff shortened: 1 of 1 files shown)


fruffy:
Hi @p2peek can you set the base to the exporter branch so we can see the actual diff?

PR #336, opened 2026-04-09
Add various folding / canonicalization patterns for `p4hir.br` and `p4hir.cond_br`

asl:


include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
```diff
@@ -2021,6 +2021,8 @@ def BrOp : P4HIR_Op<"br",
     ```
   }];
 
+  let hasCanonicalizeMethod = 0;
```
mtsamis: Shouldn't this be `let hasCanonicalizer = 1;` instead?

mtsamis (approved):
Loogs good, please see one comment.

include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
asl: Could be. Both works actually :)

mtsamis: I see :)

Well it makes sense since the patterns were applying.

PR #339, opened 2026-04-22
[P4HIR] Fix unsound CastOp chain folding (#317)

mtsamis:
Cast chains A→B→C were unconditionally folded to A→C, but this changes semantics when B alters signedness before a widening cast (e.g. bit<8>→int<8>→int<16> sign-extends, but bit<8>→int<16> zero-extends).

Only fold when the second cast doesn't widen, or the first cast preserves signedness without truncating.

Fixes #317

(This is based on https://github.com/p4lang/p4mlir-incubator/pull/333 but additionally addressing bool and bit<1> cast roundtrips).


asl (approved):
Thanks!

PR #358, opened 2026-06-05
Add extra commands to CI build to help debug differences ...

jafingerhut:
between CI system and my local system, where the build fails.

### DIFF
--- a/build_tools/build_mlir.sh
+++ b/build_tools/build_mlir.sh
@@ -22,6 +22,14 @@ LLVM_INSTALL_DIR=$P4MLIR_REPO_DIR/install
 mkdir -p "$LLVM_BUILD_DIR"
 cd "$LLVM_BUILD_DIR"
 
+# Andy's temporary extra commands to show in CI exactly what Ubuntu
+# packages are installed before the successful build below, to see if
+# that helps him learn how to get a successful build on his local
+# system.
+uname -a
+cat /etc/os-release
+cat /var/lib/dpkg/status
+
 # Configure CMake flags
 # Note that P4C uses both RTTI and C++ exceptions, so we need to build LLVM/MLIR having them enabled as well
 CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=$LLVM_INSTALL_DIR"
--- a/build_tools/build_p4c_with_p4mlir_ext.sh
+++ b/build_tools/build_p4c_with_p4mlir_ext.sh
@@ -65,7 +65,15 @@ CMAKE_FLAGS+=" -DLLVM_EXTERNAL_LIT=$LLVM_BUILD_DIR/bin/llvm-lit"
 # system.
 uname -a
 cat /etc/os-release
-cat /var/lib/dpkg/status
+#cat /var/lib/dpkg/status
+gcc --version
+g++ --version
+clang --version
+clang++ --version
+ld --version
+cmake --version
+make --version
+ninja --version
 
 # Build P4C with P4MLIR extension
 mkdir -p "$P4C_BUILD_DIR"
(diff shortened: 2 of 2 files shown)


asl:
@jafingerhut Maybe I can help with debugging things? What are the issues you're seeing?

jafingerhut:
@asl Thanks for the offer.  I have never installed p4mlir-incubator before, and was trying to do so on a local Ubuntu 24.04 Desktop Linux system, freshly installed (with at most a handful of packages over the base minimal install that the Desktop installer adds).  I think the same thing happens regardless of whether the processor is x86_64 or aarch64.

My first try was to attempt following what I believe is similar to what the Github action does in this repo:
```
git clone https://github.com/p4lang/p4mlir-incubator
cd p4mlir-incubator
git submodule update --init
./build_tools/ubuntu_install_mlir_requirements.sh |& tee log-install-try1.txt
./build_tools/build_mlir.sh |& tee log-build-mlir-try1.txt
```
but the last command above fails very quickly in what appears to be a way that is a fairly basic thing -- failing to compile what is probably a simple C++ test program to validate that one has clang++ installed correctly.  I have attached the two log files to this comment in case you are interested in taking a look.
[log-install-try1.txt](https://github.com/user-attachments/files/28657833/log-install-try1.txt)
[log-build-mlir-try1.txt](https://github.com/user-attachments/files/28657834/log-build-mlir-try1.txt)

asl:
> C++ test program to validate that one has clang++ installed correctly.

Yes, this is strange. Looks like clang was unable to find libstdc++ (C++ standard library), do not know what is wrong as it is Ubuntu-packaged one... Although, clang is not a prerequisite for p4mlir at all. Feel free to use system gcc (so, remove `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` from cmdline invocation).

jafingerhut:
Actually, a correction to my earlier comment.  If I started with an _unmodified_ Ubuntu 24.04 Desktop Linux install, I can follow the instructions to build P4MLIR successfully.

My failures were actually on systems where I had used my p4-guide script to install PI, behavioral-model, p4c, the tutorials repo, and one or two other things.  I probably will not take the time required to figure out why whatever it installs inteferes with building p4mlir-incubator, as if I want to use p4mlir-incubator I can just spin up a separate fresh Ubuntu VM for that purpose.

I ran p4mlir-translate on one of the P4-16 test programs (v1model-const-entries-bmv2.p4 was the one I chose fairly arbitrarily to try it on first) and it outputs a bunch of stuff that looks like ASCII code, probably in some kind of intermediate form that p4mlir produces by default.  Is that output able to be fed into LLVM or some other tools in order to produce x86_64 or aarch64 executable code?

asl:
> I ran p4mlir-translate on one of the P4-16 test programs (v1model-const-entries-bmv2.p4 was the one I chose fairly arbitrarily to try it on first) and it outputs a bunch of stuff that looks like ASCII code, probably in some kind of intermediate form that p4mlir produces by default. Is that output able to be fed into LLVM or some other tools in order to produce x86_64 or aarch64 executable code?

p4mlir-incubator project aims to provide an MLIR high-level representation for P4 programs in the form of dedicated MLIR Dialect (https://mlir.llvm.org/docs/LangRef/) as well as set of passes that operates on this representation (we call dialect `P4HIR`). Further translation is expected to be done by a dedicated "backend". There are unfinished eEBPF (that actually emits LLVM IR suitable for eBPF LLVM target) and BMV2 backends (aimed to produce BMV2 JSON).

jafingerhut:
Closing this PR, as I have been able to successfully build p4mlir from source code using existing scripts in the build_tools directory on a freshly installed Ubuntu 24.04 Desktop Linux system.

PR #359, opened 2026-06-15
Add canonicalization for constant struct_op

mtsamis:


lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -1796,6 +1796,24 @@ void P4HIR::StructOp::getAsmResultNames(function_ref<void(Value, StringRef)> set
     setNameFn(getResult(), name);
 }
 
+LogicalResult P4HIR::StructOp::canonicalize(P4HIR::StructOp op, PatternRewriter &rewriter) {
+    bool isCst = llvm::all_of(op.getInput(), [](mlir::Value val) {
```
asl: Can we do folding pattern here instead? As we fold `struct(const, const, ..., const)` into const. It will do const checking for us as well.

mtsamis: Yes (I incorrectly thought that fold pattern is not possible here).

mtsamis:
In latest revision:
 - Changed `struct_op` canonicalization pattern to fold.
 - Add new `llvm::ArrayRef` builder to `AggAttr`

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -1796,6 +1796,15 @@ void P4HIR::StructOp::getAsmResultNames(function_ref<void(Value, StringRef)> set
     setNameFn(getResult(), name);
 }
 
+OpFoldResult P4HIR::StructOp::fold(FoldAdaptor adaptor) {
+    // Fold extract from aggregate constant
+    bool isCstAgg =
+        llvm::all_of(adaptor.getInput(), [](mlir::Attribute attr) { return attr != nullptr; });
```
asl, line 1802:
```suggestion
        llvm::all_of(adaptor.getInput(), llvm::not_equal_to(nullptr));
```
or something like this?

asl: and I'd remove boolean variable :)

mtsamis: Looks like `llvm::not_equal_to` is not yet available in the LLVM version we're using. In that case I'd keep the extra variable due to the length of the statement.

asl:
Do we want to have similar folding for `tuple` while here?

mtsamis:
> Do we want to have similar folding for `tuple` while here?

Yes, added tuple folding and corresponding test in latest revision

asl (approved):
Thanks!

PR #361, opened 2026-06-15
Potentially canonicalize concat of shifts

mtsamis:
Try to fold concat of adjacent slices, potentially also looking through various casts.

Also introduce some useful `CastOp` helpers.

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
+++ b/include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
@@ -274,6 +274,46 @@ def CastOp : P4HIR_Op<"cast",
     `:` type($result) attr-dict
   }];
 
+  let extraClassDeclaration = [{
+    auto getSrcType() {
+      return getSrc().getType();
+    }
+
+    // Return true iff this is a cast to change the signedness of a fixed-width integer.
+    bool isSignCast() {
+      auto innerBitsType = mlir::dyn_cast<P4HIR::BitsType>(getSrcType());
+      auto resBitsType = mlir::dyn_cast<P4HIR::BitsType>(getType());
+      return innerBitsType && resBitsType &&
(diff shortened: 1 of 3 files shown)


include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
```diff
@@ -274,6 +274,42 @@ def CastOp : P4HIR_Op<"cast",
     `:` type($result) attr-dict
   }];
 
+  let extraClassDeclaration = [{
+    // Return true iff this is a cast to change the signedness of a fixed-width integer.
+    bool isSignCast() {
+      auto innerBitsType = mlir::dyn_cast<P4HIR::BitsType>(getSrc().getType());
```
asl: Let's add `getSrcType` helper to `CastOp`.

lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -724,6 +724,47 @@ OpFoldResult P4HIR::ConcatOp::fold(FoldAdaptor adaptor) {
     return P4HIR::IntAttr::get(getContext(), resultType, result);
 }
 
+LogicalResult P4HIR::ConcatOp::canonicalize(P4HIR::ConcatOp op, PatternRewriter &rewriter) {
+    // Canonicalize (A[H2:L2] ++ A[H1:L1]) to A[H2:L1] if H1 == H2.
```
asl: should it be `if H1 == L2`?

include/p4mlir/Dialect/P4HIR/P4HIR_Ops.td
mtsamis: Added in latest revision.

asl (approved):
Thanks!

PR #362, opened 2026-06-17
Implement header setInvalid

mtsamis:
Fix missing header `setInvalid` implementation. Previously we would silently generate no code.


asl (approved):
Wow, nice catch! :)

PR #363, opened 2026-06-22
Handle zero table entries

mtsamis:
Allow a zero sized region for `table_entries`. This fixes generating IR that does not verify if the user declares `(const) entries = {}`.

I didn't add a canonicalization to eliminate the empty table actions because if they're marked `const` then this encodes a constraint on the control plane side (although it wouldn't make much sense, it's still valid code).

We could eliminate it in the non-const case, but I don't see much value in that. I believe the backend is responsible to properly handle it.


asl (approved):
Can we add .p4 test as well? Into one of translator existing table tests?

mtsamis:
> Can we add .p4 test as well? Into one of translator existing table tests?

Added!

PR #366, opened 2026-07-17
Implement various Matchers for P4HIR operations

mtsamis:
Implement useful matchers for P4 operations, similar to what is available in `mlir/IR/Matchers.h`.


asl (approved):
Yay!

PR #373, opened 2026-08-14
Lower P4HIR and/or/xor, div/mod, and sadd/ssub to LLVM dialect

tagolog:


test/Conversion/P4HIRToLLVM/division-errors.mlir
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: 2026 The P4 Language Consortium
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// RUN: p4mlir-opt %s --lower-p4hir-to-llvm -split-input-file -verify-diagnostics
+
+// A divisor known to be zero is diagnosed rather than lowered into undefined behaviour.
+// p4c rejects such programs too, so this only fires on hand-written IR.
```
asl, line 8: We should never throw such errors in conversion. Conversion operates on *valid IR*, so such invariants should be diagnosed in `verify` at most.

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
```diff
@@ -86,21 +158,32 @@ struct LowerP4HIRToLLVMPass : public P4::P4MLIR::impl::LowerP4HIRToLLVMBase<Lowe
...
     }
 };
 
 }  // namespace
 
 void P4::P4MLIR::populateP4HIRToLLVMTypeConversion(LLVMTypeConverter &converter) {
-    converter.addConversion([](P4HIR::BitsType bitsType) {
+    converter.addConversion([](P4HIR::BitsType bitsType) -> std::optional<Type> {
+        // P4 allows `bit<0>`, LLVM has no `i0`: leave such values unconverted.
+        if (bitsType.getWidth() == 0) {
```
asl, line 172: LLVM's style guide suggest not to use braces on single-line bodies.

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
```diff
@@ -86,21 +158,32 @@ struct LowerP4HIRToLLVMPass : public P4::P4MLIR::impl::LowerP4HIRToLLVMBase<Lowe
...
         return IntegerType::get(bitsType.getContext(), bitsType.getWidth());
     });
 
     converter.addTypeAttributeConversion(
-        [&converter](P4HIR::BitsType bitsType, P4HIR::IntAttr attr) {
-            return IntegerAttr::get(converter.convertType(bitsType), attr.getValue());
+        [&converter](P4HIR::BitsType bitsType,
+                     P4HIR::IntAttr attr) -> LLVMTypeConverter::AttributeConversionResult {
+            // Types without an LLVM counterpart (e.g. `bit<0>`) have no attribute either.
+            auto convertedType = converter.convertType(bitsType);
```
asl, line 182: why not write something like this:
```c++
if (auto convertedType = converter.convertType(bitsType)) {
  return IntegerAttr::get(convertedType, attr.getValue());
}

return LLVMTypeConverter::AttributeConversionResult::na();
```

This looks cleaner IMO.

test/Conversion/P4HIRToLLVM/division-errors.mlir
tagolog: Fixed.

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
tagolog: Dropped the braces. .clang-tidy wants them, but it is not enforced (same as p4c).

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
tagolog: Fixed.

asl (approved):

PR #374, opened 2026-08-14
Lower P4HIR bool, unary and cmp to LLVM dialect

tagolog:
Stacked on #373, review the last commit only. Will rebase once #373 lands.

### DIFF
--- a/lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
+++ b/lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
@@ -99,6 +99,90 @@ struct BinOpConversion : public ConvertOpToLLVMPattern<P4HIR::BinOp> {
     }
 };
 
+struct UnaryOpConversion : public ConvertOpToLLVMPattern<P4HIR::UnaryOp> {
+    using ConvertOpToLLVMPattern::ConvertOpToLLVMPattern;
+
+    LogicalResult matchAndRewrite(P4HIR::UnaryOp op, OpAdaptor adaptor,
+                                  ConversionPatternRewriter &rewriter) const override {
+        auto input = adaptor.getInput();
+        auto intType = dyn_cast<IntegerType>(input.getType());
+        if (!intType) {
+            return rewriter.notifyMatchFailure(op, "unsupported unary operand type");
+        }
+
+        auto createConstant = [&](int64_t value) {
+            return LLVM::ConstantOp::create(rewriter, op.getLoc(), intType, value);
+        };
+
+        switch (op.getKind()) {
+            case P4HIR::UnaryOpKind::UPlus:
+                rewriter.replaceOp(op, input);
+                return success();
+            case P4HIR::UnaryOpKind::Neg:
+                // LLVM has no integer negation, so `-x` is emitted as `0 - x`.
+                rewriter.replaceOpWithNewOp<LLVM::SubOp>(op, createConstant(0), input);
+                return success();
+            case P4HIR::UnaryOpKind::Cmpl:
+                rewriter.replaceOpWithNewOp<LLVM::XOrOp>(op, input, createConstant(-1));
+                return success();
+            case P4HIR::UnaryOpKind::LNot:
+                if (intType.getWidth() != 1) {
+                    return rewriter.notifyMatchFailure(op, "logical not requires i1 operand");
+                }
+                rewriter.replaceOpWithNewOp<LLVM::XOrOp>(op, input, createConstant(1));
+                return success();
+        }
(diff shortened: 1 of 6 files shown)


lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
```diff
@@ -99,6 +99,90 @@ struct BinOpConversion : public ConvertOpToLLVMPattern<P4HIR::BinOp> {
...
 
+struct UnaryOpConversion : public ConvertOpToLLVMPattern<P4HIR::UnaryOp> {
+    using ConvertOpToLLVMPattern::ConvertOpToLLVMPattern;
+
+    LogicalResult matchAndRewrite(P4HIR::UnaryOp op, OpAdaptor adaptor,
+                                  ConversionPatternRewriter &rewriter) const override {
+        auto input = adaptor.getInput();
+        auto intType = dyn_cast<IntegerType>(input.getType());
+        if (!intType) {
+            return rewriter.notifyMatchFailure(op, "unsupported unary operand type");
```
asl, line 110: Can it really be so? Or it should be just explicit `cast<>` above?

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
```diff
@@ -99,6 +99,90 @@ struct BinOpConversion : public ConvertOpToLLVMPattern<P4HIR::BinOp> {
...
+            return LLVM::ConstantOp::create(rewriter, op.getLoc(), intType, value);
+        };
+
+        switch (op.getKind()) {
+            case P4HIR::UnaryOpKind::UPlus:
+                rewriter.replaceOp(op, input);
+                return success();
+            case P4HIR::UnaryOpKind::Neg:
+                // LLVM has no integer negation, so `-x` is emitted as `0 - x`.
+                rewriter.replaceOpWithNewOp<LLVM::SubOp>(op, createConstant(0), input);
```
asl, line 123: You used to have helper for this, no?

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
```diff
@@ -99,6 +99,90 @@ struct BinOpConversion : public ConvertOpToLLVMPattern<P4HIR::BinOp> {
...
+        }
+
+        bool isSigned = false;  // BoolType lowers to i1 and is always compared as unsigned.
+        if (auto bitsType = dyn_cast<P4HIR::BitsType>(lhsType)) {
+            isSigned = bitsType.isSigned();
+        }
+
+        switch (op.getKind()) {
+            case P4HIR::CmpOpKind::Eq:
+                return lowerToICmpOp(op, adaptor, rewriter, LLVM::ICmpPredicate::eq);
```
asl, line 162: can you do lambda to capture operands?

asl:
@tagolog please rebase

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
tagolog: Fixed

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
tagolog: Replaced with helper call

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
tagolog: Changed to the lambda call

lib/Conversion/P4HIRToLLVM/LowerP4HIRToLLVM.cpp
```diff
@@ -117,6 +122,74 @@ struct BinOpConversion : public ConvertOpToLLVMPattern<P4HIR::BinOp> {
...
+        };
+
+        switch (op.getKind()) {
+            case P4HIR::UnaryOpKind::UPlus:
+                rewriter.replaceOp(op, input);
+                return success();
+            case P4HIR::UnaryOpKind::Neg:  // `-x` is emitted as `0 - x`
+                return lowerToOp<LLVM::SubOp>(op, {createConstant(0), input}, rewriter);
+            case P4HIR::UnaryOpKind::Cmpl:  // `~x` is emitted as `x ^ -1`
+                return lowerToOp<LLVM::XOrOp>(op, {input, createConstant(-1)}, rewriter);
```
chbessonova, line 144: Would this still be correct for widths above 64?
If I read this right, `intType` is already signless here, so the `-1` mask would be
zero-extended.

PR #375, opened 2026-08-18
Implement IndexableType, IndexedField and FieldPath

mtsamis:
This PR consists of:
 - A new helper class `FieldPath` to model an arbitrary field access path within a type with fields and provide useful methods to operate on such paths.
 - A simplification / unification of our type-with-fields interfaces.

Some of the rationale and design choice for the interface refactoring:
 - We were missing a "this type can be indexed into" interface that can reasonably cover both Arrays and  Struct-like types. This is especially important for a `FieldPath` implementation, since without that we wouldn't be able to model a path of an element in an array within a struct, which is a totally normal P4 construct. As such, implementing any relevant transformation (e.g. struct flattening) would be problematic or limited.
 - The previous `FieldIDTypeInterface` was complicated to implement and too low-level to be useful. It didn't expose the most commonly needed functionality, so it most code would just cast to StructLikeType or ArrayType instead.

Under this new implementation:
 - `IndexableTypeInterface` is trivially easy to implement and each provided function is easy enough to understand that it is self-ocumenting.
 - `IndexedField` is a new struct that represents the field access within an `IndexableTypeInterface`. It is essentially just a wrapper that conveniently provides the `IndexableTypeInterface` functionality for a particular field. Crucially, it is also meant to represent an array element (or any user-defined indexable type results). This is meant to replace almost all uses of `FieldInfo`.
 - `FieldInfo` is now only meant to be used to construct StructLike types (so I'm considering renaming it to `StructFieldInfo`).
 - `StructLikeTypeInterface` is now a thin wrapper on top of `IndexableTypeInterface`. I've kept it so that code can still easily check for struct-like types, wherever that makes sense. It also has couple additional methods.
 - The fieldID related infrastructure is now implemented separately within `FieldPath`. We only need an indexable-type and we can implement all fieldID calculations without user code, since the most important primitive that needs to exist is `getMaxFieldID` and that can be recursively computed by calling `getFieldCount()`.

### DIFF
--- a/include/p4mlir/Dialect/P4HIR/FieldPath.h
+++ b/include/p4mlir/Dialect/P4HIR/FieldPath.h
@@ -1,0 +1,145 @@
+// SPDX-FileCopyrightText: 2025 The P4 Language Consortium
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef P4MLIR_DIALECT_P4HIR_P4HIR_FIELDPATH_H
+#define P4MLIR_DIALECT_P4HIR_P4HIR_FIELDPATH_H
+
+#include "p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.h"
+
+namespace P4::P4MLIR::P4HIR {
+
+/// Represents a field path starting from a root type.
+struct FieldPath {
+ private:
+    explicit FieldPath(mlir::Type rootType, mlir::Type leafType, unsigned fieldID)
+        : rootType(rootType), leafType(leafType), fieldID(fieldID) {}
+
+ public:
+    /// Get the max fieldID that reference a field of `type`.
+    static unsigned getMaxFieldID(mlir::Type type);
+
+    /// Get the fieldID for the field at `index` in `itype`.
+    static unsigned getFieldID(P4HIR::IndexableTypeInterface itype, unsigned index);
+
+    /// The results of indexing in an IndexableType using a fieldID.
+    struct IndexingResult {
+        IndexingResult(unsigned index, unsigned parentFieldId, unsigned childFieldId)
+            : index(index), parentFieldId(parentFieldId), childFieldId(childFieldId) {}
+
+        /// The index of the field in the IndexableType type that corresponds to the given fieldID.
+        unsigned index;
+        /// The fieldID for the field at `index`.
+        unsigned parentFieldId;
+        /// The sub-fieldID to index within the resulting field.
+        unsigned childFieldId;
+    };
+
+    /// Find the immidiate field of `itype` that contains the field referenced by `fieldID` and
+    /// return indexing information.
+    static IndexingResult indexInto(P4HIR::IndexableTypeInterface itype, unsigned fieldID);
+
+    /// Create a FieldPath by fully indexing into `rootType` using `fieldID`.
+    static FieldPath fromFieldID(mlir::Type rootType, unsigned fieldID);
+
+    /// Construct a special empty FieldPath.
+    explicit FieldPath() : FieldPath(mlir::Type(), mlir::Type(), 0) {}
+
+    /// Construct a field path representing `rootType`.
+    explicit FieldPath(mlir::Type rootType) : FieldPath(rootType, rootType, 0) {}
+
+    /// Returns true if the given fields paths can be concatenated into a single path with `concat`.
+    static bool canConcat(const FieldPath &lhs, const FieldPath &rhs) {
+        return lhs.isEmpty() || rhs.isEmpty() || lhs.getLeafType() == rhs.getRootType();
+    }
+
+    /// Given two fields paths X->...->Y and Y->...->Z returns X->...->Z.
+    /// Returns FieldPath() if either pah is the special empty path.
+    /// Assumes that `canConcat(lhs, rhs) == true`.
+    static FieldPath concat(const FieldPath &lhs, const FieldPath &rhs);
+
+    /// If this path is of the form X->...->Y->...->Z where Y is the first field for
+    /// which `pred(Y)` is true then return two new field paths X->...->Y and Y->...->Z.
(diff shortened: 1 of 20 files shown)


include/p4mlir/Dialect/P4HIR/FieldPath.h
```diff
@@ -0,0 +1,145 @@
...
+
+        /// The index of the field in the IndexableType type that corresponds to the given fieldID.
+        unsigned index;
+        /// The fieldID for the field at `index`.
+        unsigned parentFieldId;
+        /// The sub-fieldID to index within the resulting field.
+        unsigned childFieldId;
+    };
+
+    /// Find the immidiate field of `itype` that contains the field referenced by `fieldID` and
```
asl:
```suggestion
    /// Find the immediate field of `itype` that contains the field referenced by `fieldID` and
```

include/p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.h
```diff
@@ -5,41 +5,50 @@
...
 
-/// Struct defining a field. Used in structs and header
+/// Struct that represents a particular field of an IndexableType.
+struct IndexedField {
+    IndexedField(mlir::Type indexableType, unsigned index)
+        : indexableType(indexableType), index(index) {}
+    unsigned getIndex() const { return index; }
+    mlir::Type getParentType() const { return indexableType; }
+    mlir::Type getType() const;
+    std::string getName() const;
```
asl, line 21: It should really return `StringRef`.

include/p4mlir/Dialect/P4HIR/P4HIR_Types.td
```diff
@@ -411,117 +411,33 @@ class StructLikeType<string name, string typeMnemonic>
...
+      return getElements()[index].annotations;
     }
 
-    std::optional<unsigned> getFieldIndex(mlir::StringAttr fieldName) {
-      llvm::ArrayRef<P4HIR::FieldInfo> elems = getElements();
-      for (size_t idx = 0, numElems = elems.size(); idx < numElems; ++idx)
-        if (elems[idx].name == fieldName) return idx;
+    std::optional<P4HIR::IndexedField> getFieldByName(mlir::StringRef fieldName) {
+      for (auto field : getFields())
+        if (field.getName() == fieldName) return field;
```
asl: This creates temporary string on each comparison

mtsamis: Thanks, changed to `getNameAttr`

include/p4mlir/Dialect/P4HIR/P4HIR_TypeInterfaces.h
mtsamis: That's a trade-off on whether we want to allow on-the-fly generated names for types implementing these interfaces, which I used for arrays.

The problem with a `StringRef` variant would be then either
1) This class cannot be used for ArrayType
2) We have to remove `getName` from there, which is annoying when it comes to struct-type.

If you think we should change it to `StringRef` (and hence remove it from that struct), maybe one approach would be to subclass this into `NamedIndexedField` and use that subclass for struct-like-type-interface.

asl: Ok, given that other place now uses `getNameAttr()` we should be fine with keeping this as-is.

mtsamis:
In the latest (major) revision:
 - Implement various iterators to provide easier access, replace `forEach` methods.
 - Add many new helpers and new functionality.
 - Seperate static fieldID related helpers in `FieldIDs` struct.
 - Fix bugs from initial implementation, clean up code.

include/p4mlir/Dialect/P4HIR/FieldPath.h
```diff
@@ -0,0 +1,327 @@
...
+        unsigned rhsFid;
+    };
+
+    PathIterator begin() const { return PathIterator(getRootType(), getFieldID()); }
+    PathIterator end() const { return PathIterator::end(); }
+
+ public:
+    /// Iterate all prefix paths that make up this path.
+    /// For a path T.x.y.z this will yield T, T.x, T.x.y and T.x.y.z.
+    auto iterPaths() const {
```
asl: I'd name it using STL-like notation, so `paths` and `fields`.

lib/Dialect/P4HIR/FieldPath.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    FieldPath newPath(newRootType);
+    for (auto field : iterFields()) newPath.append(field.getIndex());
+
+    return newPath;
+}
+
+std::string FieldPath::getIdentifier(llvm::StringRef delimiter) const {
+    llvm::SmallString<64> name;
+    for (auto field : iterFields()) {
```
asl: `llvm::interleave` here?

lib/Dialect/P4HIR/FieldPath.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    return false;
+}
+
+FieldPath FieldPath::withRoot(mlir::Type newRootType) const {
+    if (isEmpty()) return *this;
+
+    FieldPath newPath(newRootType);
+    for (auto field : iterFields()) newPath.append(field.getIndex());
```
asl: looks like `llvm::append_range(newPath, llvm::map_range(fileds(), []{...})`. I'd add `field_indices()` helper that would essentially do `map_range`.

mtsamis: Added `field_indices` as a helper, but `append_range` cannot be done since it requires `C.insert(C.end(), adl_begin(R), adl_end(R));`. I don't think such an `insert` function would make a lot of sense to exist for `FieldPath`.

asl (approved):

PR #377, opened 2026-09-11
Implement new PathWalker utility

mtsamis:
Implement a new `FieldPath`-based utility to walk field-access users of values.


include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+
+        auto lookThroughArray = lookThroughArrayAccess(op, operand, path);
+        if (lookThroughArray.first) return lookThroughStruct;
+
+        return {mlir::Value(), P4HIR::FieldPath()};
+    }
+
+    /// Walk field-access operations and uses starting from `root`. If `rootType` is additionally
+    /// provided it is used to compute field paths instead of `root.getType()`. This is useful when
+    /// doing type convertions. If `rootType` is provided it must have the same shape as
```
asl:
```suggestion
    /// doing type conversions. If `rootType` is provided it must have the same shape as
```

include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+        indirectUses = indirectUsesMap;
+        return *this;
+    }
+
+    /// Set the callback function for leaf operations. This callback is called for operations that
+    /// are not field-access operations and that access a value with a known path. The operation,
+    /// the operand and the corresponding path are provided as arguments to the callback. If a leaf
+    /// operation has multiple operands with known paths then the callback will be called multiple
+    /// times.
+    PathWalker &setLeafCallback(LeafCallbackFn cb) {
```
asl, line 145: Maybe add constructor that allows to specify callbacks and two static methods – one for leaf-only and another one – for node-only callbacks being set?

include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+    }
+
+    /// Set the indirect use mapping to use. This mapping provides additional uses for values that
+    /// are accessed by means other than SSA def-use chains (e.g. symbol-based access).
+    PathWalker &setIndirectUsesMap(const IndirectUsesMap *indirectUsesMap) {
+        indirectUses = indirectUsesMap;
+        return *this;
+    }
+
+    /// Set the callback function for leaf operations. This callback is called for operations that
```
asl, line 140: I would annotate types / ctors instead :)

include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+                                                                           mlir::OpOperand &operand,
+                                                                           P4HIR::FieldPath path) {
+        return llvm::TypeSwitch<mlir::Operation *, std::pair<mlir::Value, P4HIR::FieldPath>>(op)
+            .Case<P4HIR::ArrayGetOp, P4HIR::ArrayElementRefOp>(
+                [&](auto arrayAccessOp) -> std::pair<mlir::Value, P4HIR::FieldPath> {
+                    assert((operand.get() == arrayAccessOp.getInput()) && "Unexpected operand");
+
+                    unsigned cstIndex;
+                    if (matchPattern(arrayAccessOp.getIndex(), m_ConstantInt(&cstIndex)))
+                        return {arrayAccessOp.getResult(), path[cstIndex]};
```
asl, line 181: Do we want to distinguish between "non-constant array access" vs "unknown access"? Do we need to have some field path placeholder for variable indices?

include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+    /// operations. The result value and the corresponding accessed field are provided as arguments
+    /// to the callback.
+    PathWalker &setNodeCallback(NodeCallbackFn cb) {
+        nodeCb = std::move(cb);
+        return *this;
+    }
+
+    /// Helper to look through a P4HIR `struct_extract` or `struct_field_ref` operation.
+    static std::pair<mlir::Value, P4HIR::FieldPath> lookThroughStructAccess(
+        mlir::Operation *op, mlir::OpOperand &operand, P4HIR::FieldPath path) {
```
asl, line 160: don't we want to pass `operand` here by const reference? We're not modifying it, and `.get()` is declared as `const`.

include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+                assert((operand.get() == structAccessOp.getInput()) && "Unexpected operand");
+                return std::pair{structAccessOp.getResult(), path[structAccessOp.getFieldIndex()]};
+            })
+            .Default(
+                [](mlir::Operation *) { return std::pair{mlir::Value(), P4HIR::FieldPath()}; });
+    }
+
+    /// Helper to look through a P4HIR `array_get` or `array_element_ref` operation.
+    static std::pair<mlir::Value, P4HIR::FieldPath> lookThroughArrayAccess(mlir::Operation *op,
+                                                                           mlir::OpOperand &operand,
```
asl, line 172: ditto

include/p4mlir/Transforms/IRUtils.h
```diff
@@ -102,6 +105,152 @@ class SplitStateRewriter {
...
+            .Default(
+                [](mlir::Operation *) { return std::pair{mlir::Value(), P4HIR::FieldPath()}; });
+    }
+
+    /// `operand` is an operand in `op` that is used to access the root value's field that
+    /// corresponds to `path`. If `op` is a field access operation and the resulting path can be
+    /// determined based on `path`, then this function should return the resulting new value and
+    /// path. Otherwise an empty value must be returned.
+    virtual std::pair<mlir::Value, P4HIR::FieldPath> lookThroughUser(mlir::Operation *op,
+                                                                     mlir::OpOperand &operand,
```
asl, line 194: ditto

include/p4mlir/Transforms/IRUtils.h
mtsamis: I have thought about this a bit; it feels to me like `FieldPath` shouldn't support variable indices.

`FieldPath` is a fieldID-based wrapper after all, whereas arbitrary access with unknowns is much more heavyweight and different to handle. I think they're mostly separate things.

mtsamis:
Changed implementation to use constructor and static `forNodes`/`forLeafs` methods.

asl (approved):

PR #379, opened 2026-09-11
Improve canonicalization and folding of boolean and validity bit comp…

mtsamis:
Adds a number of comparison and validity-bit related folding.

Validity bits are special:
They don't have integer representation through P4HIR `getConstantInt` and they also require an additional comparison to get a bool from. The added patterns help to to fold a wide variety of `isValid` usage patterns among other things.

(Draft until tests are added)


lib/Dialect/P4HIR/P4HIR_Ops.cpp
```diff
@@ -923,6 +935,86 @@ OpFoldResult P4HIR::CmpOp::fold(FoldAdaptor adaptor) {
...
+            if (lhs != op.getLhs() || rhs != op.getRhs()) {
+                auto newKind = newKindIsEq ? P4HIR::CmpOpKind::Eq : P4HIR::CmpOpKind::Ne;
+                rewriter.replaceOpWithNewOp<P4HIR::CmpOp>(op, newKind, lhs, rhs);
+                return success();
+            }
+        }
+
+        // Helper to check if `val` is a validity bit check, equivalent to (validBit ==
+        // valid/invalid).
+        auto matchValidityCheck = [](mlir::Value val) -> std::pair<mlir::Value, bool> {
```
asl, line 964: Can we have `isValid` / `isNotValid` common pattern matchers instead, please? :)

mtsamis: Yes, done!

asl (approved):

PR #381, opened 2026-09-21
p4hir.for and p4hir.foreach lowering

PietroGhg:


lib/Transforms/FlattenCFG.cpp
```diff
@@ -128,15 +128,66 @@ class ScopeOpFlattening : public mlir::OpRewritePattern<P4HIR::ScopeOp> {
     }
 };
 
+struct ForOpFlattening : public OpRewritePattern<P4HIR::ForOp> {
```
asl, line 134: Can we also have `ForIn` lowering as well?

